PR #1, opened 2022-10-05
Message handling(private, normal, commands), sit down and stand up ad…

Cr3ativeZen:
.


apps/gameserver/include/Character.h
```diff
@@ -24,6 +24,10 @@ class Character : public bango::space::quad_entity
     int             m_z;
     std::uint8_t    m_map;
 
+    std::uint16_t   m_n64GState;
+    std::uint16_t   m_n64MState;
+    std::uint16_t   m_n64GStateEx;
+    std::uint16_t   m_n64MStateEx;
```
lafreak, line 30: Those are ROAM conventions (adding `n` before names and UpperCase names)
Stick to the same naming convention as above (m_z/m_map)

Also, the data types seem wrong - only 2 bytes? If I remember correctly those should be 8 bytes and signed (std::int64_t)

```suggestion
    std::int64_t   m_gstate;
    std::int64_t   m_mstate;
    std::int64_t   m_gstate_ex;
    std::int64_t   m_mstate_ex;
```

apps/gameserver/include/World.h
```diff
@@ -218,6 +218,18 @@ class World
         for (auto& p : Get().m_entities[Character::PLAYER])
             callback((Player*) p.second);
     }
+    static const Player* FindPlayerByName(const std::string& playerName)
+    {
+        std::lock_guard<std::recursive_mutex> lock(Get().m_entities_rmtx);
+        for (auto& p : Get().m_entities[Character::PLAYER])
+        {
+            auto* player = dynamic_cast<const Player*>(p.second);
+            if(player and playerName == player->GetName())
```
lafreak, line 227: Use `&&` instead of `and`.
You could also use std::find and utilize c++11 features: https://en.cppreference.com/w/cpp/algorithm/find instead of writing for loop

apps/gameserver/src/Player.cpp
```diff
@@ -135,21 +135,70 @@ void Player::OnLoadFinish()
...
-    std::cout << "On Rest " << (int)action << std::endl;
+    packet restPacket(S2C_ACTION);
+    restPacket << GetID() << AT_REST << action;
+	if (action)
+	{
+		if (IsGState(CGS_REST))
+			return;
+
+		AddGState(CGS_REST);
+        World::Map(GetMap()).WriteInSight(this, restPacket);
```
lafreak, line 146: Tab is missing

apps/gameserver/src/Player.cpp
```diff
@@ -135,21 +135,70 @@ void Player::OnLoadFinish()
...
+    switch(message.front())
+    {
+        case '/':
+        {
+            CommandDispatcher::Dispatch(this, message);
+            break;
+        }
+        case '@':
+        {
+            std::string receiverName = message.substr(1, message.find(' ') - 1);
```
lafreak, line 177: What if there is no space in the message? `"@hello"`

apps/gameserver/src/Player.cpp
```diff
@@ -135,21 +135,70 @@ void Player::OnLoadFinish()
...
+
+            if(receiverName != GetName())
```
lafreak, line 179: You can exit earlier here and not create new level of indentation, just:
```
if (receiver_name == GetName())
  return;
```

apps/gameserver/src/Player.cpp
```diff
@@ -135,21 +135,70 @@ void Player::OnLoadFinish()
...
+                if(receiver)
+                {
+                    write(messagePacket);
+                    receiver->write(messagePacket);
+                }
+                else
+                {
+                    packet infoPacket(S2C_MESSAGE);
+                    infoPacket << MSG_THEREISNOPLAYER;
+                    write(infoPacket);
```
lafreak, line 191: To save lines you could use another `write` overload: `write(S2C_MESSAGE, "d", MSG_THEREISNOPLAYER);`
There is also issue with pushing `MSG_THEREISNOPLAYER` directly into packet because you cannot be sure what type it is. By specifying "d" you can be sure it is 4 bytes.

apps/gameserver/src/Player.cpp
```diff
@@ -135,21 +135,70 @@ void Player::OnLoadFinish()
...
+        {
+            std::string receiverName = message.substr(1, message.find(' ') - 1);
+
+            if(receiverName != GetName())
+            {
+                auto* receiver = World::FindPlayerByName(receiverName);
+                if(receiver)
+                {
+                    write(messagePacket);
+                    receiver->write(messagePacket);
```
lafreak, line 185: Writer should get the message as well, right? This is missing

apps/gameserver/src/Player.cpp
```diff
@@ -135,21 +135,70 @@ void Player::OnLoadFinish()
...
-        return CommandDispatcher::Dispatch(this, message);
+    auto message = p.pop_str();
 
-    packet out(S2C_CHATTING);
-    out << GetName() << message;
+    packet messagePacket(S2C_CHATTING);
+    messagePacket << GetName() << message;
 
-    World::Map(GetMap()).WriteInSight(this, out);
+    switch(message.front())
```
lafreak, line 168: You removed the comment that empty string can crash, but did not fix the issue. Stl specifies that calling `front()` on empty string is undefined behavior:
https://cplusplus.com/reference/string/string/front/

apps/gameserver/src/Player.cpp
Cr3ativeZen: It was just looking for a player but didnt send a message. I will add check for it so it will not look a player.

apps/gameserver/include/Character.h
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: the check at the beggining if(p.empty) is not enough?

apps/gameserver/include/Character.h
Cr3ativeZen: done

apps/gameserver/include/World.h
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: I've changed it to .at(0), but as we discussed on discord, checking if comming packet is empty solves the problem.

apps/gameserver/src/Player.cpp
lafreak: If space will not be found, it will evaluate to `std::npos - 1` which is undefined behavior.
We should check if message contains space before proceeding.
```
auto space_pos = message.find(' ');
if (space_pos == std::npos)
  return;
std::string receiver_name = message.substr(1, space_pos);
```

apps/gameserver/src/Player.cpp
lafreak: Checking if packet is empty is other than checking if message is empty.
The first one can be non-empty but the other one can be empty in the same time.
How?

Imagine such packet:
```
0x03 0x00 0x?? 0x00
```
0x?? - C2S_CHATTING

Packet is not empty - it contains single byte of data: 0x00 (first 3 bytes are size and type).
Then we pop string:
```
auto message = p.pop_str();
```
Message is now empty.
How so? In the packet, the strings are null terminated strings: https://en.wikipedia.org/wiki/Null-terminated_string
It means that "ABC" is 0x41 0x42 0x43 0x00
Empty string "" is 0x00

In summary, the code should include:
```
if (message.empty())
  return;
```

apps/gameserver/include/Character.h
```diff
@@ -84,6 +88,24 @@ class Character : public bango::space::quad_entity
     std::uint64_t   GetGStateEx()   const { return 0; }
     std::uint64_t   GetMStateEx()   const { return 0; }
```
lafreak, line 89: Line 86-89 should be updated to reflect gstate changes in appear packets.
Currently it always returns 0.

```
std::uint64_t GetGState() const { return m_state; }
etc
```

apps/gameserver/include/Character.h
```diff
@@ -84,6 +88,24 @@ class Character : public bango::space::quad_entity
...
 
+    void AddGState(std::int64_t gstate)       { m_gstate |= gstate; }
+    void AddMState(std::int64_t mstate)       { m_mstate |= mstate; }
+
+    void AddGStateEx(std::int64_t gstateEx)   { m_gstate_ex |= gstateEx; }
+    void AddMStateEx(std::int64_t mstateEx)   { m_mstate_ex |= mstateEx; }
```
lafreak, line 95: We should stick to one convention.
This is good start: https://google.github.io/styleguide/cppguide.html

The variables should be in snake_case.
Class members too, with m_ prefix.
Class methods should be in UpperCase.

apps/gameserver/include/Character.h
lafreak: What's more we can see that we send uint64_t as gstate in appear packet, so it is good idea to decide to go with `std::uint64_t` instead of `std::int64_t` for states:
```
void AddGState(std::uint64_t gstate) { m_gstate |= gstate; }
```

apps/gameserver/include/World.h
```diff
@@ -218,6 +218,20 @@ class World
...
+        std::lock_guard<std::recursive_mutex> lock(Get().m_entities_rmtx);
+        const auto& players = Get().m_entities[Character::PLAYER];
+
+        auto playerIterator = std::find_if(players.begin(),players.end(), [&](auto& p){
+            auto* player = dynamic_cast<const Player*>(p.second);
+
+            return player && playerName == player->GetName();
+        });
+
+        return playerIterator != players.end() ? dynamic_cast<const Player*>(playerIterator->second) : nullptr;
```
lafreak, line 232: Prefer reinternet_cast over dynamic_cast, dynamic cast adds overhead in runtime

lafreak: https://stackoverflow.com/questions/332030/when-should-static-cast-dynamic-cast-const-cast-and-reinterpret-cast-be-used

lafreak (approved):

PR #6, opened 2022-10-12
Monster die v2

Cr3ativeZen:


apps/gameserver/include/BeheadableMonster.h
```diff
@@ -0,0 +1,16 @@
...
+
+class BeheadableMonster : public Monster
+{
+public:
+    BeheadableMonster(const std::unique_ptr<InitMonster>& init, int x, int y, int map=0) : Monster(init, x, y) {}
+    void Die() override;
+    void Tick() override;
+
+private:
+    std::chrono::steady_clock::time_point m_death_time;
```
lafreak, line 15: Use bango::utils::time, or is there any reason to use chrono directly here?

PR #7, opened 2022-10-12
Monster die v2

Cr3ativeZen:


apps/gameserver/include/BeheadableMonster.h
```diff
@@ -0,0 +1,16 @@
...
+
+class BeheadableMonster : public Monster
+{
+public:
+    BeheadableMonster(const std::unique_ptr<InitMonster>& init, int x, int y, int map=0) : Monster(init, x, y) {}
+    void Die() override;
+    void Tick() override;
+
+private:
+    std::chrono::steady_clock::time_point m_death_time;
```
lafreak, line 15: Use bango::utils::time, or is there any reason to use chrono directly here?

apps/gameserver/include/World.h
```diff
@@ -172,6 +172,19 @@ class World
...
 
+    static void RemoveDeadMonsters()
+    {
+        auto& monsters = Get().m_entities[Character::MONSTER];
+        for(auto it = monsters.begin(); it != monsters.end(); ++it)
+        {
+            if(it->second->IsGState(CGS_KO))
+            {
+                delete it->second;
+                monsters.erase(it++);
```
lafreak, line 183: Erase from map is not thread safe.
Other threads may use monsters container and this thread will invalidate it which is undefined behavior in cpp.

lafreak, line 183: What's more, erasing from container while iterating it also invalidates the iterator.
Check how to erase it properly: https://stackoverflow.com/questions/4713131/removing-item-from-vector-while-iterating

apps/gameserver/include/World.h
```diff
@@ -172,6 +172,19 @@ class World
...
     }
 
+    static void RemoveDeadMonsters()
+    {
+        auto& monsters = Get().m_entities[Character::MONSTER];
+        for(auto it = monsters.begin(); it != monsters.end(); ++it)
+        {
+            if(it->second->IsGState(CGS_KO))
+            {
+                delete it->second;
```
lafreak, line 182: Destructing the object is unsafe here since other threads may still use the objects, we should wait for release guard implementation IMO

apps/gameserver/src/Player.cpp
```diff
@@ -1,7 +1,7 @@
 #include "Player.h"
-
 #include "Socket.h"
 #include "World.h"
+#include "BeheadableMonster.h"
```
lafreak, line 4: I don't think we need this specific include here, `Monster.h` should be sufficient?

apps/gameserver/src/Player.cpp
```diff
@@ -614,12 +613,6 @@ void Player::OnAttack(packet& p)
         auto duration = (now-m_last_attack).count();
         auto damage_reduce = (double) duration / (double) GetAttackSpeed();
 
-        if (damage_reduce > 1.f)
```
lafreak, line 617: Why do we remove it?

apps/gameserver/src/Player.cpp
```diff
@@ -632,21 +625,43 @@ void Player::OnAttack(packet& p)
...
+        else if (damage_reduce < 0.6f)
+            return;
+
         std::int64_t damage = GetAttack();
         damage *= damage_reduce;
         // Something with mage
         damage = character->GetFinalDamage(this, damage);
         // Apply Mix Effects
 
+        //BUG: It is possible to do more damage than remaining HP of target.
```
lafreak, line 640: The bug can be easily fixed by checking first if damage is larger than current hp.
In this case we could reduce the damage to equal current hp.

```
damage = std::min(damage, character->GetCurHP());
World::Map(GetMap()).WriteInSight ...
character->ReduceHP(damage);
if (character->GetCurHP() <= 0)
  character->Die();
```

apps/gameserver/src/main.cpp
```diff
@@ -106,7 +108,24 @@ int main(int argc, char** argv)
         int index = token;
 
         try {
-            World::Add(new Monster(InitMonster::DB().at(index), player->GetX()+10, player->GetY()+10, player->GetMap()));
+            switch(InitMonster::DB().at(index)->Race)
```
lafreak, line 111: Create static method CreateMonster that builds monster based on index, x, y and map.
The logic will be reused in area spawner and boss spawners.

apps/gameserver/include/BeheadableMonster.h
Cr3ativeZen: I've changed it to bango::utils::time.

apps/gameserver/include/World.h
Cr3ativeZen: I've changed it to remove_if, lets push it for now. You will fix it with the release guard

apps/gameserver/include/World.h
Cr3ativeZen: as above.

Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/main.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: if (character->GetCurHP() <= 0)
GetCurHP() is unsigned int

Cr3ativeZen: and what i meant is that packet sends wrong amount when killing(if mob has 133 hp, and u can deal 600, server sends packet with 600 damage not 133, thats how it should work).

Cr3ativeZen: anyway its done

apps/gameserver/src/Player.cpp
```diff
@@ -632,21 +625,42 @@ void Player::OnAttack(packet& p)
...
         World::Map(GetMap()).WriteInSight(this, packet(S2C_ATTACK, "ddddb", 
             GetID(), 
             character->GetID(),
```
lafreak, line 643:
```cpp
        std::int64_t damage = GetAttack();
        damage *= damage_reduce;
        // Something with mage
        damage = character->GetFinalDamage(this, damage);
        // Apply Mix Effects

        if (damage < 0)  // first ensure the data is unsigned.
            return;
        if ((uint64_t)damage > character->GetCurHP())  // then we can safely cast to unsigned64
            damage = character->GetCurHP();

        World::Map(GetMap()).WriteInSight(this, packet(S2C_ATTACK, "ddddb", 
            GetID(), 
            character->GetID(), 
            damage,
            0,//EB
            damage == 0 ? ATF_IGNORE : ATF_HIT));

        character->ReduceHP(damage);  // inside reducehp we could even throw an exception to ensure we never have less than 0 hp

        if(character->GetCurHP() <= 0)
            character->Die();

```

lafreak (approved):

PR #12, opened 2022-10-25
Monster spawn

Cr3ativeZen:


apps/gameserver/include/Character.h
```diff
@@ -13,14 +13,15 @@ class Character : public bango::space::quad_entity
 {
 public:
     typedef std::uint32_t id_t;
+    static id_t g_max_id;
```
lafreak, line 16: This should not be public. ID generation should be hidden inside Character class and not exposed to other classes.

lafreak: To achieve it, I suggest creating new public method: `AssignNewId` which in future will be threadsafe (my task) and will be used inside Character constructor and spawner.

apps/gameserver/include/Monster.h
```diff
@@ -74,11 +75,15 @@ struct InitMonster : public bango::processor::db_object<InitMonster>
 class Monster : public Character
 {
     const std::unique_ptr<InitMonster>& m_init;
+    bango::utils::time::point m_death_time;
 
 public:
     Monster(const std::unique_ptr<InitMonster>& init, int x, int y, int map=0);
     ~Monster();
 
+    void PrepareMonsterToSpawn(int new_x, int new_y, int map);
```
lafreak, line 84:
```suggestion
    void Respawn(int new_x, int new_y, int map);
```

apps/gameserver/include/Spawn.h
```diff
@@ -0,0 +1,55 @@
...
+struct GenMonster : public bango::processor::db_object<GenMonster>
+{
+    struct RectXY
+    {
+        int X1 = 0, Y1 = 0, X2 = 0, Y2 = 0;
+        int GetRandomX() const;
+        int GetRandomY() const;
+
+    };
+    std::uint32_t MonsterIndex = 0, Map = 0, Index = 0, Amount = 0, SpawnCycle = 0;
```
lafreak, line 24:
```suggestion
    std::uint32_t MonsterIndex = 0, Map = 0, Area= 0, Amount = 0, SpawnCycle = 0;
```
Why not call it as it is inside config file?

apps/gameserver/include/World.h
```diff
@@ -129,10 +130,12 @@ class World
     typedef std::unordered_map<Character::id_t, Player*>                    PlayerContainer;
     typedef std::unordered_map<Character::id_t, std::shared_ptr<Monster>>   MonsterContainer;
     typedef std::unordered_map<Character::id_t, std::shared_ptr<NPC>>       NpcContainer;
+    typedef std::unordered_map<Character::id_t, std::shared_ptr<Spawn>>     SpawnContainer;
```
lafreak, line 133: Is there any reason why we should keep it inside a map?
I don't think we will ever need to identify spawn by id.
It does not inherit from Character.
Could it be std::list or any container which is not key value?

apps/gameserver/include/World.h
```diff
@@ -179,12 +182,15 @@ class World
             Add(std::make_shared<NPC>(init.second));
     }
 
+    static void SpawnGenMonster();
```
lafreak, line 185:
```suggestion
    static void SpawnMonsters();
```

apps/gameserver/include/World.h
```diff
@@ -179,12 +182,15 @@ class World
...
+
     static void Cleanup()
     {
         std::lock_guard<std::recursive_mutex> lock(Get().m_entities_rmtx);
         Get().m_players.clear();
         Get().m_monsters.clear();
         Get().m_npcs.clear();
+        Get().m_spawns.clear();
```
lafreak, line 193: Here we remove monsters before we remove spawns. Is it safe? Will spawner keep respawning monsters because monsters got deleted? What is the expected behavior here?

apps/gameserver/src/Monster.cpp
```diff
@@ -63,24 +63,48 @@ packet Monster::BuildMovePacket(std::int8_t delta_x, std::int8_t delta_y, std::i
...
-            World::Add(std::make_shared<RegularMonster>(InitMonster::DB().at(index), x, y, map));
+            World::Add(std::make_shared<RegularMonster>(init_monster, x, y, map));
             break;
         }
     }
+}
+
+void Monster::SetNewID()
+{
+    m_id = Character::g_max_id++;
```
lafreak, line 91: Why moved declaration out of this function? I think we should minimize access to that variable and only have access to it here.

apps/gameserver/src/Monster.cpp
```diff
@@ -63,24 +63,48 @@ packet Monster::BuildMovePacket(std::int8_t delta_x, std::int8_t delta_y, std::i
...
+}
+
+void Monster::PrepareMonsterToSpawn(int new_x, int new_y, int map)
```
lafreak, line 94:
```suggestion
void Monster::RestoreInitialState(int new_x, int new_y, int map)
```

lafreak: Why is `int map` here?

apps/gameserver/src/RegularMonster.cpp
```diff
@@ -2,13 +2,15 @@
...
 #include <bango/network/packet.h>
+#include <bango/utils/time.h>
 #include <inix.h>
 
 using namespace bango::network;
 
 void RegularMonster::Die()
 {
     SetGState(CGS_KO);
+    SetDeathTime(bango::utils::time::now());
```
lafreak, line 13: Why do RegularMonsters need to have death time set?

apps/gameserver/src/Spawn.cpp
```diff
@@ -0,0 +1,157 @@
...
+}
+
+void Spawn::Tick()
+{
+    if((time::now() - m_next_spawn_cycle).count() > GetSpawnCycle())
+        return;
+
+    for(auto& monster : m_area_monsters)
+    {
+        auto monster_lock = monster->Lock();
```
lafreak, line 52: Here you lock because you want to check if monster is ready to be respawned (CGS_KO).
I don't think you really need to lock.

In the context of monster _just_ got killed and you ask too quick for CGS_KO, its not a problem. We will not respawn but the next tick iteration will respawn it, right? We can live with that.

In the context of we checking CGS_KO and we are afraid that when we perform respawn some other thread will change state (remove CGS_KO) - can it really happen? I think the only thread that can remove CGS_KO should be this tick thread during respawn. Other thread should NEVER do that, right?

We need to ensure no other thread edits monster during respawn. How to do that?
Which other threads can edit? Player recv packets. For example player->monster attack.
If we ensure that monster is already CGS_KO during attack packet we SHOULD reject attack. This will ensure us that nothing can write to it beside spawner.

To summarize, CGS_KO can be our safe flag that there is only one thread that can edit monster - Ticker.

apps/gameserver/src/Spawn.cpp
```diff
@@ -0,0 +1,157 @@
...
+    World::Add(monster);
+}
+
+void Spawn::CreateSpawn()
+{
+    const auto& init_monster = InitMonster::DB().at(GetMonsterIndex());
+    m_area_monsters.resize(GetAmount());
+    for (int i = 0; i < GetAmount(); i++)
+    {
+        switch (init_monster->Race)
```
lafreak, line 72: This switch duplicates switch inside Monster::Summon. This is not maintainable in the long run - what if we add 5 more monster kinds - do we want to remember to add to support all types in both places. I dont think we need to.

My suggestion is to separate the Monster builder method and spawn method.

Monster::CreateMonster could be the builder and create monster based on init,x,y,map. Create but not add to the world. We could have the switch in there.

Other logic should be separated and re-use Monster::CreateMonster.

Here we could call Monster::CreateMonster, check if it is not nullptr (error during creation) and insert into the world and spawner container.
Other place (on demand Summon - /mob or boss spawn) could also reuse Monster::CreateMonster and add to the world in separate Monster::Summon method.

apps/gameserver/src/World.cpp
```diff
@@ -326,4 +329,21 @@ void WorldMap::WriteOnMap(const packet& p)
 
     for (const auto& pair : Players())
         ((Player*)pair.second)->write(p);
+}
+
+void World::SpawnGenMonster()
+{
+    for(const auto& init : GenMonster::DB())
+    {
+        auto new_spawn = std::make_shared<Spawn>(GenMonster::DB().at(init.second->Index));
```
lafreak, line 338: Here we iterate over GenMonster entries (`init` variable containing key and value pair (`area` and `GenMonster` ptr)).
For each entry, we retrieve the area (`init.second->Index`) and use it to get ` GenMonster` ptr. Why? We already have the ptr inside `init` variable. `init.second` is the pointer.

apps/gameserver/src/World.cpp
```diff
@@ -326,4 +329,21 @@ void WorldMap::WriteOnMap(const packet& p)
...
+        Get().m_spawns.insert(std::make_pair(init.second->Index, new_spawn));
+    }
+}
+
+void World::ForEachSpawn(const std::function<void(Spawn&)>& callback)
+{
+    std::lock_guard<std::recursive_mutex> lock(Get().m_entities_rmtx);
```
lafreak, line 345: We currently do not need to lock. The container does not change in runtime.
It will probably change if we add hot reload at runtime (inix supports that?)

apps/gameserver/include/Character.h
Cr3ativeZen: done

apps/gameserver/include/Spawn.h
Cr3ativeZen: done

apps/gameserver/src/World.cpp
Cr3ativeZen: ups, done

apps/gameserver/include/World.h
Cr3ativeZen: done

apps/gameserver/include/World.h
Cr3ativeZen: done

apps/gameserver/src/Monster.cpp
Cr3ativeZen: done

apps/gameserver/src/Monster.cpp
Cr3ativeZen: done

apps/gameserver/src/World.cpp
Cr3ativeZen: Inix does not support that i think, but it's an useful feature and we should have it later on. I will remove lock_guard for now.

apps/gameserver/src/RegularMonster.cpp
Cr3ativeZen: done

core/processor/include/bango/processor/db.h
```diff
@@ -5,6 +5,7 @@
 #include <map>
 #include <functional>
```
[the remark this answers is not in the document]
Cr3ativeZen: ups, done

apps/gameserver/src/Spawn.cpp
Cr3ativeZen: reload at spawn time might be an issue later on.
i will remove lock for now

apps/gameserver/include/World.h
Cr3ativeZen: It might be a case, and mob could get added via World::Add back to m_monsters.

apps/gameserver/src/Spawn.cpp
Cr3ativeZen: done

apps/gameserver/src/Monster.cpp
```diff
@@ -63,48 +63,30 @@ packet Monster::BuildMovePacket(std::int8_t delta_x, std::int8_t delta_y, std::i
     return packet();
 }
 
-void Monster::SummonMonster(std::uint32_t index, std::int32_t x, std::int32_t y, std::int32_t map)
+void Monster::Summon(std::uint32_t index, std::int32_t x, std::int32_t y, std::int32_t map)
+{
+    World::Add(Monster::CreateMonster(index, x, y, map));
```
lafreak, line 68: Error handling is missing - CreateMonster can return nullptr

apps/gameserver/include/Character.h
```diff
@@ -39,15 +39,15 @@ class Character : public bango::space::quad_entity
...
-        m_id = g_max_id++;
+        AssignNewId();
     }
 
     constexpr static std::uint8_t PLAYER    =0;
     constexpr static std::uint8_t MONSTER   =1;
     constexpr static std::uint8_t NPC       =2;
     constexpr static std::uint8_t LOOT      =3;
 
+    void AssignNewId();
```
lafreak, line 50: Could this be protected? So Monster class can use it but no other classes can.

apps/gameserver/include/World.h
```diff
@@ -179,10 +182,13 @@ class World
             Add(std::make_shared<NPC>(init.second));
     }
 
+    static void SpawnMonster();
```
lafreak, line 185:
```suggestion
    static void SpawnMonsters();
```

apps/gameserver/src/Spawn.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+void Spawn::Tick()
+{
+    if((time::now() - m_next_spawn_cycle).count() > GetSpawnCycle())
+        return;
+
+    for(auto& monster : m_area_monsters)
+    {
+        if(monster->IsGState(CGS_KO))
```
lafreak, line 52: Add note here that we have thread-safety by making sure we always call RemoveDeadMonsters and Spawn::Tick in the same thread and that Spawn::Tick is called **after** RemoveDeadMonsters.

apps/gameserver/src/Spawn.cpp
```diff
@@ -0,0 +1,137 @@
...
+            RespawnOnWorld(monster);
+    }
+
+    SetNextSpawnCycle();
+}
+
+void Spawn::RespawnOnWorld(std::shared_ptr<Monster> monster)
+{
+    monster->RestoreInitialState(GetRandomX(), GetRandomY());
+    World::Add(monster);
```
lafreak, line 62: World add/remove API is missing check if the monster with given ID already exist. It's not an issue here but in the future it will be important to catch bugs early. Add TODO.

apps/gameserver/include/Spawn.h
```diff
@@ -0,0 +1,55 @@
...
+
+class Spawn
+{
+public:
+
+    Spawn(const std::unique_ptr<GenMonster>& init);
+
+    void Tick();
+private:
+    void RespawnOnWorld(std::shared_ptr<Monster> monster);
```
lafreak, line 39: This can be const

apps/gameserver/src/Spawn.cpp
lafreak, line 52: This call can crash if we end up in line 72.

```

    for (int i = 0; i < GetAmount(); i++)
    {
        auto monster = Monster::CreateMonster(GetMonsterIndex(), GetRandomX(), GetRandomY(), GetMap());
        if(!monster)
            spdlog::error("Monster was not created correctly.");
        m_area_monsters.at(i) = monster;  // HERE WE POSSIBLY PUT nullptr
        World::Add(monster);
    }
```

and when we put that nullptr and we iterate over vector here in tick we will dereference nullptr.

apps/gameserver/src/Spawn.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+void Spawn::CreateSpawn()
+{
+    const auto& init_monster = InitMonster::DB().at(GetMonsterIndex());
+    m_area_monsters.resize(GetAmount());
+    for (int i = 0; i < GetAmount(); i++)
+    {
+        auto monster = Monster::CreateMonster(GetMonsterIndex(), GetRandomX(), GetRandomY(), GetMap());
```
lafreak, line 71: As far I see the transaction policy for CreateMonster method is that we always return valid ptr. The only exit from this function is by throwing an exception. We should use that and catch that exception instead of calling `if(!monster)`

lafreak: We should not add into the list if the exception happen

lafreak: We also should not add nullptr to the world

apps/gameserver/include/World.h
lafreak: Even SpawnMonsters is misleading in this case.
This is responsible for two things:
- creating spawns (the containers)
- spawns monsters related to those spawns

My final suggestion is: `CreateSpawns`

lafreak: Or even better: `CreateSpawnsAndSpawnMonsters`

apps/gameserver/src/Spawn.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+using namespace bango::utils;
+
+Spawn::Spawn(const std::unique_ptr<GenMonster>& init)
+    :m_init(init)
+{
+    if(InitMonster::DB().find(GetMonsterIndex()) != InitMonster::DB().end())
+        CreateSpawn();
+    else
+        spdlog::error("Monster ID does not exist in InitMonster {}", GetMonsterIndex()); // TODO: object shouldn't be created.
```
lafreak, line 21: This can easily be done in World::CreateSpawnsAndSpawnMonsters method.

```
void World::CreateSpawnsAndSpawnMonsters()
{
    for(const auto& init : GenMonster::DB())
    {
        if (InitMonster::DB().count(init.second->MonsterIndex) == 0)
        {
             log error;
             continue;
         }
        Get().m_spawns.emplace_back(std::make_shared<Spawn>(init.second));
    }
}
```

lafreak: Then we can get rid of the if in line 18.

lafreak (changes requested):
You are 5 commits away from master. Please rebase your changes.

apps/gameserver/include/World.h
Cr3ativeZen: done

apps/gameserver/include/Character.h
Cr3ativeZen: done

apps/gameserver/src/Spawn.cpp
Cr3ativeZen: done

apps/gameserver/src/Spawn.cpp
Cr3ativeZen: done

apps/gameserver/include/Spawn.h
Cr3ativeZen: done

apps/gameserver/src/Spawn.cpp
Cr3ativeZen: done

apps/gameserver/src/Spawn.cpp
Cr3ativeZen: done

Cr3ativeZen: done

apps/gameserver/src/Spawn.cpp
Cr3ativeZen: try catch added instead, Monster::CreateMonster aka std::make_shared will never return nullptr.

apps/gameserver/src/Spawn.cpp
lafreak: Add proper formatting (add indent inside try block)

Do not use resize and at - you are still able to leave nullptr in the m_area_monsters and then dereference it in Spawn::Tick which is unsafe

apps/gameserver/src/World.cpp
```diff
@@ -326,4 +330,25 @@ void WorldMap::WriteOnMap(const packet& p)
...
         ((Player*)pair.second)->write(p);
+}
+
+void World::CreateSpawnsAndSpawnMonsters()
+{
+    for(const auto& init : GenMonster::DB())
+    {
+        if (InitMonster::DB().count(init.second->MonsterIndex) == 0)
+        {
+            spdlog::error("Monster ID does not exist in InitMonster {}", init.second->MonsterIndex);
```
lafreak, line 341: Add spawn id (area) into the log as well for better debugging

Cr3ativeZen: done

apps/gameserver/src/Spawn.cpp
Cr3ativeZen: my bad...
done

lafreak (approved):

PR #21, opened 2022-11-06
Exp distribution

Cr3ativeZen:


apps/gameserver/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 
 project(gameserver)
 
-set (CMAKE_CXX_STANDARD 14)
+set (CMAKE_CXX_STANDARD 20)
```
lafreak, line 5: I think you do not use any cpp20 feature, but cpp17 (for range loop). Do not require higher standard that we actually use.

apps/gameserver/src/Monster.cpp
```diff
@@ -89,4 +90,40 @@ void Monster::RestoreInitialState(int new_x, int new_y)
...
+    std::map<id_t, std::uint64_t> party_container;
+    for (auto&[id, damage] : hostility_map)
+    {
+        World::ForPlayer(id, [&](Player& player){
+            auto player_lock = player.Lock();
+            if (player.GetPartyID() != 0)
+                party_container[player.GetPartyID()] += damage;
+            else
+            {
+                std::uint64_t exp = static_cast<double>(damage) / all_hostility * m_init->Exp;
```
lafreak, line 114: Casting damage which is uint64_t to double can cause precision loss.
https://stackoverflow.com/questions/63093880/divide-two-64-bit-numbers-and-store-the-result-in-a-float-without-losing-precisi

apps/gameserver/include/Player.h
```diff
@@ -145,14 +146,19 @@ class Player : public Character, public User
...
+    id_t GetPartyID()                                   const   { return m_party_id; }
 
     std::uint16_t   GetReqPU(std::uint8_t* stats);
 
     void Tick() override;
     void Die() override;
+
+    void UpdateExp(std::int64_t amount);
+    bool CanReciveExp();
+    void CalculateExp(std::uint64_t& exp, std::uint8_t monster_level);
```
lafreak, line 163: Why passing the reference? Can we return the result instead?

apps/gameserver/src/Party.cpp
```diff
@@ -104,8 +110,25 @@ void Party::RemoveMember(Player* player, bool is_kicked)
         SendPartyInfo();
         if (was_leaver_party_leader)
             GetLeader()->write(S2C_MESSAGE, "b", MSG_BECOMEPARTYHEAD);
+        if (was_leaver_party_top_level)
+            SetTopLevel();
```
lafreak, line 114: Getters and setters should be prefixed with Get and Set. Here we prefix with Set but in fact this is not a setter (no value is passed as first parameter). I suggest to rename to `UpdateTopLevel`

apps/gameserver/src/Party.cpp
```diff
@@ -69,7 +74,7 @@ bool Party::AddMember(Player* player)
         p.push<unsigned short>(player->GetMaxHP());
         WriteToAll(p);
     }
-
+    m_top_level = player->GetLevel() > m_top_level ? player->GetLevel() : m_top_level;
```
lafreak, line 77: What do you think about calling `UpdateTopLevel` here? I know current implementation is more performance (we do not need to iterate over player list) but we could encapsulate setting top level behavior in one single place. I'm not sure what's better.

apps/gameserver/src/Player.cpp
```diff
@@ -679,7 +679,8 @@ void Player::OnAttack(packet& p)
             0,//EB
             damage == 0 ? ATF_IGNORE : ATF_HIT));
```
lafreak, line 681: We can actually unlock the attacker earlier, when we are sure not to modify or use player's dynamic properties any more.
I think this is line 668, right?

apps/gameserver/src/Player.cpp
```diff
@@ -824,4 +827,39 @@ void Player::Die()
 {
     AddGState(CGS_KO);
     WriteInSight(bango::network::packet(S2C_ACTION, "db", GetID(), AT_DIE));
+}
+
+void Player::UpdateExp(std::int64_t amount)
+{
+    if(GetLevel() >= 100) // ExpTable ends at lvl 100
+        return;
```
lafreak, line 835: We should inform administrator why the exp is not being added. spdlog warn?

apps/gameserver/src/Player.cpp
```diff
@@ -824,4 +827,39 @@ void Player::Die()
...
+    m_data.Exp += amount;
+
+    write(S2C_UPDATEPROPERTY, "bII", P_EXP, GetExp(), amount);
+}
+
+bool Player::CanReciveExp()
+{
+    return !IsGState(0x0000023); // check for 3 gstates: CGS_KNEE, CGS_KO, CGS_FISH
```
lafreak, line 843: 0x23 is magic constant here. Use existing enum and combine all 3:
`CGS_KNEE | CGS_KO | CGS_FISH`

apps/gameserver/src/Player.cpp
```diff
@@ -824,4 +827,39 @@ void Player::Die()
...
+}
+
+void Player::CalculateExp(std::uint64_t& exp, std::uint8_t monster_level)
+{
+    //TODO: Calculate exp buffs like exp stone, asadal, exp event.
+    std::int8_t level_difference =  monster_level - GetLevel();
+
+    if (level_difference < 0)
+    {
+        if (std::abs(level_difference) >= 20)
```
lafreak, line 853: you can use `std::min`:
```
level_difference = std::min(std::abs(level_difference), 20);
```

apps/gameserver/src/Player.cpp
```diff
@@ -824,4 +827,39 @@ void Player::Die()
...
+            level_difference = 20;
+        else
+            level_difference = std::abs(level_difference);
+        exp = (exp - exp * (g_nReviseExpB[(GetLevel() - 1) / 10][level_difference] / 100.0));
```
lafreak, line 857: Calculating monster color and applying ratio based on color ratio is in one line here. It is also duplicated in 863. Separate those operations and add coments which step does what. It should help us maintain the calculations.

lafreak (changes requested):

apps/gameserver/include/Monster.h
```diff
@@ -75,12 +76,16 @@ struct InitMonster : public bango::processor::db_object<InitMonster>
 class Monster : public Character
 {
     const std::unique_ptr<InitMonster>& m_init;
+    std::map<id_t, std::uint64_t> hostility_map;
+    std::uint64_t all_hostility;
+    void ExpDistribution();
```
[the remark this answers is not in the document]
Cr3ativeZen: done

apps/gameserver/include/Monster.h
```diff
@@ -75,12 +76,16 @@ struct InitMonster : public bango::processor::db_object<InitMonster>
 class Monster : public Character
 {
     const std::unique_ptr<InitMonster>& m_init;
+    std::map<id_t, std::uint64_t> hostility_map;
+    std::uint64_t all_hostility;
```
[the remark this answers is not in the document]
Cr3ativeZen: I think it's easier to add damage to hostility_map every time monster gets hit because suming all values in hostility_map every monster death might be time consuming when more players hits same mob. Mainsrv has this value aswell.

apps/gameserver/include/Monster.h
```diff
@@ -75,12 +76,16 @@ struct InitMonster : public bango::processor::db_object<InitMonster>
 class Monster : public Character
 {
     const std::unique_ptr<InitMonster>& m_init;
+    std::map<id_t, std::uint64_t> hostility_map;
```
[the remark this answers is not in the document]
Cr3ativeZen: no we dont, done

apps/gameserver/include/Character.h
```diff
@@ -84,7 +84,7 @@ class Character : public bango::space::quad_entity
     std::uint32_t   GetCurHP()  const { return m_curhp; }
     std::uint32_t   GetCurMP()  const { return m_curmp; }
 
-    void ReduceHP(std::uint32_t reduce);
+    virtual void ReceiveDamage(id_t id, std::uint32_t damage);
```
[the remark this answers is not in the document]
Cr3ativeZen: Only id is needed in this case

apps/gameserver/include/Party.h
```diff
@@ -17,14 +18,20 @@ class Party
 {
     std::list<Player*> m_members_list;
     mutable std::recursive_mutex m_rmtx_list;
+    id_t m_id;
+    std::uint8_t m_top_level{0};
 public:
     static constexpr std::uint8_t MAX_PARTY_SIZE = 8;
     static constexpr std::uint8_t MIN_PARTY_SIZE = 2;
+    static constexpr std::uint16_t EXP_RECEIVE_RANGE = 640;
```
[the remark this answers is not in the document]
Cr3ativeZen: it is range from monster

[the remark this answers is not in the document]
Cr3ativeZen: > Is it range from monster or from party leader? Or party center?
done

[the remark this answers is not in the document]
Cr3ativeZen: > Is it range from monster or from party leader? Or party center?

done

apps/gameserver/src/Player.cpp
Cr3ativeZen: yes, done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: line 857(g_nReviseExpB) calculates exp when Monster has lower level than player.
line 863(g_nReviseExpA) calculates exp when Monster has higher level than player.
I added some comments

apps/gameserver/include/Monster.h
```diff
@@ -2,6 +2,7 @@
 
 #include <cstdint>
 #include <memory>
+#include <map>
```
lafreak, line 5: Not used, include unordered_map

apps/gameserver/CMakeLists.txt
Cr3ativeZen: done

apps/gameserver/src/Monster.cpp
Cr3ativeZen: done

apps/gameserver/include/Player.h
Cr3ativeZen: done

apps/gameserver/src/Party.cpp
Cr3ativeZen: well, since party creation is not used so often we might aswell use UpdateTopLevel here.

apps/gameserver/include/Monster.h
Cr3ativeZen: ups,done

apps/gameserver/src/Party.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
```diff
@@ -850,16 +854,15 @@ void Player::CalculateExp(std::uint64_t& exp, std::uint8_t monster_level)
...
-            level_difference = 20;
-        else
-            level_difference = std::abs(level_difference);
+        level_difference = std::min(std::abs(level_difference), 20);
+        //When monster_level < player_level, use g_nReviseExpB to calculate color ratio
         exp = (exp - exp * (g_nReviseExpB[(GetLevel() - 1) / 10][level_difference] / 100.0));
     }
     else
     {
         if(level_difference >= 20)
```
lafreak, line 864: Use std min here as well. This can be unified to single call before if/else.

apps/gameserver/src/Monster.cpp
```diff
@@ -111,9 +111,8 @@ void Monster::DistributeExp()
                 party_container[player.GetPartyID()] += damage;
             else
             {
-                std::uint64_t exp = static_cast<double>(damage) / total_hostility * m_init->Exp;
-                player.CalculateExp(exp, GetLevel());
-                player.UpdateExp(exp);
+                std::uint64_t exp = static_cast<long double>(damage) / total_hostility * m_init->Exp;
```
lafreak, line 114: To be sure about casting order first calculate share percentage:
```
double exp_share = static_cast<double>(damage) / static_cast<double>(total_hostility);
```
and then use it to get share out of total exp:
```
exp *= exp_share;
```

apps/gameserver/src/Party.cpp
```diff
@@ -74,7 +74,7 @@ bool Party::AddMember(Player* player)
         p.push<unsigned short>(player->GetMaxHP());
         WriteToAll(p);
     }
-    m_top_level = player->GetLevel() > m_top_level ? player->GetLevel() : m_top_level;
+    UpdateTopLevel();
```
lafreak, line 77: I don't think it will work if you call it before adding player to the list. Have you tested it?

apps/gameserver/src/Player.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Player.h"
 
 #include <string>
+#include <algorithm>
```
lafreak, line 4: Why do we need it?

apps/gameserver/include/Party.h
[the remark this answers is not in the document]
lafreak: > it is range from monster

If so, why is it in Party header? It can be misleading

apps/gameserver/src/Monster.cpp
Cr3ativeZen: done

apps/gameserver/src/Party.cpp
Cr3ativeZen: yyes it wont work ... ups, done

apps/gameserver/src/Player.cpp
Cr3ativeZen: we dont, done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
```diff
@@ -850,21 +849,17 @@ bool Player::CanReciveExp()
 std::uint64_t Player::CalculateExp(std::uint64_t exp, std::uint8_t monster_level)
 {
     //TODO: Calculate exp buffs like exp stone, asadal, exp event.
-    std::int8_t level_difference =  monster_level - GetLevel();
-
+    std::int32_t level_difference =  monster_level - GetLevel();
```
lafreak, line 852: This can overflow if monster_level is lower than player level since both are unsigned.

Cr3ativeZen: done

lafreak (dismissed):

lafreak (approved):

PR #25, opened 2022-11-08
Level up

Cr3ativeZen:


apps/gameserver/src/Player.cpp
```diff
@@ -829,16 +833,30 @@ void Player::Die()
...
+    if (amount == 0)
+        return;
+
+    if (GetLevel() >= 100)
     {
         spdlog::warn("ExpTable ends at lvl 100, exp will not be increased.");
         return;
     }
-    m_data.Exp += amount;
+    std::uint64_t exp_needed_for_next_level = g_nNeedExp[GetLevel()] * g_fNeedExpRatio[GetLevel()] / EXP_RATE;
```
lafreak, line 846: It would be better to do this calculation beforehand, at the server start and then re-use constant array here.

apps/gameserver/src/Player.cpp
```diff
@@ -829,16 +833,30 @@ void Player::Die()
     WriteInSight(bango::network::packet(S2C_ACTION, "db", GetID(), AT_DIE));
 }
 
-void Player::UpdateExp(std::int64_t amount)
+void Player::UpdateExp(std::int64_t amount, bool send_update_to_client)
 {
-    if(GetLevel() >= 100) // ExpTable ends at lvl 100
+    if (amount == 0)
```
lafreak, line 838: What if amount is lower than 0?

apps/gameserver/src/Player.cpp
```diff
@@ -829,16 +833,30 @@ void Player::Die()
...
+        return;
+
+    if (GetLevel() >= 100)
```
lafreak, line 841: We are currently not possible to decrease EXP if we are 100+ lv. Is that desired?

apps/gameserver/src/Player.cpp
```diff
@@ -829,16 +833,30 @@ void Player::Die()
...
     {
         spdlog::warn("ExpTable ends at lvl 100, exp will not be increased.");
         return;
     }
-    m_data.Exp += amount;
+    std::uint64_t exp_needed_for_next_level = g_nNeedExp[GetLevel()] * g_fNeedExpRatio[GetLevel()] / EXP_RATE;
+    if (m_data.Exp + amount >= exp_needed_for_next_level)
```
lafreak, line 847: It should be possible to have 10 bars (100% of required exp) and still not level up. I think we should have `>` here?

apps/gameserver/src/Player.cpp
```diff
@@ -829,16 +833,30 @@ void Player::Die()
...
+    {
+        std::int64_t leftover_exp = m_data.Exp + amount - exp_needed_for_next_level;
+        LevelUp();
+
+        if(leftover_exp > 0)
+            UpdateExp(leftover_exp, false);
+    }
+    else
+        m_data.Exp += amount;
```
lafreak, line 856: Correct me if I'm wrong but if we send update exp packet here then we do not need bool parameter in this function.

apps/gameserver/src/Player.cpp
```diff
@@ -862,4 +880,33 @@ std::uint64_t Player::CalculateExp(std::uint64_t exp, std::uint8_t monster_level
...
+    m_data.SUPoint++;
+    m_data.PUPoint += GetAmountOfPUPointsOnLevelUp();
+    m_data.Exp = 0;
+    UpdatePropertyOnLevelUp();
+}
+
+std::uint8_t Player::GetAmountOfPUPointsOnLevelUp()
+{
+        return (
+            (GetLevel() >= 96 ? 12 :
```
lafreak, line 897: Where is this formula from? Are you sure it is not already defined in inix.h in a form of an array?

apps/gameserver/src/Player.cpp
```diff
@@ -862,4 +880,33 @@ std::uint64_t Player::CalculateExp(std::uint64_t exp, std::uint8_t monster_level
...
+void Player::LevelUp()
+{
+    m_data.Level++;
+    m_data.SUPoint++;
+    m_data.PUPoint += GetAmountOfPUPointsOnLevelUp();
+    m_data.Exp = 0;
+    UpdatePropertyOnLevelUp();
+}
+
+std::uint8_t Player::GetAmountOfPUPointsOnLevelUp()
```
lafreak, line 894: Can be const

apps/gameserver/src/Player.cpp
```diff
@@ -862,4 +880,33 @@ std::uint64_t Player::CalculateExp(std::uint64_t exp, std::uint8_t monster_level
...
+            (GetLevel() >= 72 ? 7 :
+                                5)))))));
+}
+
+void Player::UpdatePropertyOnLevelUp()
+{
+    SendProperty(P_LEVEL);
+    SendProperty(P_SUPOINT);
+    SendProperty(P_PUPOINT);
+    SaveAllProperty();
```
lafreak, line 911: I personally would prefer to have `SaveAllProperty` called outside of this function. We should avoid hiding database write/read in deeper parts of implementation.

apps/gameserver/include/Player.h
```diff
@@ -158,7 +158,11 @@ class Player : public Character, public User
     void Tick() override;
     void Die() override;
 
-    void UpdateExp(std::int64_t amount);
+    void UpdateExp(std::int64_t amount, bool send_update_to_client = true);
     bool CanReciveExp();
     std::uint64_t CalculateExp(std::uint64_t exp, std::uint8_t monster_level);
+    void LevelUp();
+private:
+    std::uint8_t GetAmountOfPUPointsOnLevelUp();
```
lafreak, line 166: This can be defined outside of Player class. I think it would not bloat player implementation. Please look up FIND_PU macro for example

apps/gameserver/include/Player.h
```diff
@@ -158,7 +158,11 @@ class Player : public Character, public User
...
+    void UpdatePropertyOnLevelUp();
```
lafreak, line 167: We are starting to get plenty of different Update methods. I think we will quickly end up in tens or hundreds of those in different flavors. Inix made it nicely with single UpdateProperty method which takes `kind` as parameter and `amount`. This way you can implement switch for all kind cases. This would not only increase/decrease the value but also take care of sending packet to the player. It would be nicely coupled for later use.

apps/gameserver/src/Player.cpp
Cr3ativeZen: I found it in r3vo code. Seems to be right imo.

apps/gameserver/src/Player.cpp
Cr3ativeZen: If we dont use Boolean here you will receive increase exp msg multiple times if u lvl up, which is not the case in Regular files.

apps/gameserver/src/Player.cpp
Cr3ativeZen: There’s nothing to decrease if u can’t increase.

apps/gameserver/src/Player.cpp
Cr3ativeZen: This Is the check if updateexp== 0 is sent. It will not do anything beside printing “ exp increased by 0”, if value is less than 0 then exp will decrease. Method name is “updateexp” not increase exp. We will use it later on to decrease exp for example after we click start from village.

apps/gameserver/src/Player.cpp
Cr3ativeZen: It’s how it’s done in inix files.

apps/gameserver/src/Player.cpp
lafreak: Entire gameserver is lacking unit tests and that is starting to bite us. There is something to decrease when you have 20 EXP left to lv100, kill monster and getp 500k exp. Then you are Lv. 100 with 499980 EXP. And you could decrease the exp in theory.

apps/gameserver/src/Player.cpp
lafreak: Then maybe we should separate the message from actual calculation and increasement?

apps/gameserver/src/Player.cpp
lafreak: To be verified (that source cant be trusted)

apps/gameserver/src/Player.cpp
Cr3ativeZen: When u have 20 exp left to 100, then get 500k exp then u will get ur exp increased by 20 and rest of exp won't cuz line 850 call(UpdateExp) is after lvl up, so you are trying to update exp for lvl 100 player, which will pop the warning. The only problem is that you will get your exp increased clientside, and after relog it will be gone.

apps/gameserver/src/Player.cpp
Cr3ativeZen: yep, done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
```diff
@@ -862,4 +884,12 @@ std::uint64_t Player::CalculateExp(std::uint64_t exp, std::uint8_t monster_level
     //When monster_level > player_level, use g_nReviseExpA to calculate color ratio
     exp = (exp * (g_nReviseExpA[(GetLevel() - 1) / 10][level_difference] / 100.0) + exp);
     return exp;
-}
\ No newline at end of file
+}
+
+void Player::LevelUp()
+{
+    SendProperty(P_LEVEL, 1);
```
lafreak, line 891: What I meant with SendProperty was a little bit different. Look up how inix does it.

I would leave SendProperty to do one thing and do it good. **Only** send property to the client. Then no _amount_ parameter is needed to the function.

I would use this function in new, `UpdateProperty` method which takes _kind_ and _amount_. The UpdateProperty increase by amount would call SendProperty without amount parameter (player already has it increased).

apps/shared/inix/common.h
```diff
@@ -29,7 +32,111 @@ static double g_fNeedExpRatio[] = {
 	5,5,5,5,5,5,5,5,5,5,5,5								//99-110
 };
 
-static std::uint8_t g_nReviseExpA[10][21] = {
+
+static constexpr std::uint64_t g_exp_table[MAX_LEVEL] = {
```
lafreak, line 36: Cannot stay like that :weary:
I'd rather invest time writing proper compiletime calculation or use C style :( global variable with extern functionality.

lafreak: https://stackoverflow.com/a/57214093/5480703

apps/shared/inix/common.h
```diff
@@ -155,6 +263,23 @@ static int g_nAddDefLv[] = {
 	99, 99, 99, 99, 99, 99, 99, 99, 99, 99, 99
 };
 
+static std::uint8_t GET_PU_ON_LEVEL_UP(std::uint8_t level)
```
lafreak, line 266: This function will be statically compiled into binary X times, where X is the number of cpp compile units (cpp files) which use this common.h header (I guess all use it).
I suggest to write macro, smilarly how FIND_PU_POINT does it.

apps/gameserver/src/Player.cpp
```diff
@@ -829,16 +838,29 @@ void Player::Die()
...
 {
-    if(GetLevel() >= 100) // ExpTable ends at lvl 100
+    if (amount == 0)
+        return false;
+
+    if (GetLevel() >= MAX_LEVEL)
     {
         spdlog::warn("ExpTable ends at lvl 100, exp will not be increased.");
-        return;
+        return false;
```
lafreak, line 849: Log states that exp will not be increased, but this function is not only to increase but decrease as well

apps/gameserver/src/Player.cpp
```diff
@@ -829,16 +838,29 @@ void Player::Die()
...
+    }
+
+    if (m_data.Exp + amount > GET_EXP_FOR_LEVEL(GetLevel()))
```
lafreak, line 852: This if will not work and overflow in case of exp decrease. This shows that we got plenty of corner cases we do not cover "by hand". I think unit tests are required here.

lafreak: In case we loose more exp than we have, this will give us 4kkkk exp = 100lv

apps/gameserver/src/Player.cpp
lafreak: Yes, I meant what is happening if exp is _lower_ than 0, not equal. The behavior is not tested, as stated above

lafreak (approved):

apps/gameserver/src/Player.cpp
lafreak: fixed

apps/shared/inix/common.h
lafreak: fixed

apps/gameserver/src/Player.cpp
lafreak: fixed

apps/gameserver/include/Player.h
lafreak: fixed

apps/gameserver/include/Player.h
lafreak: fixed

PR #29, opened 2023-02-24
Loot parsing utilities

Cr3ativeZen:


apps/gameserver/include/Loot.h
```diff
@@ -0,0 +1,83 @@
...
+};
+
+
+struct LootGroup : public bango::processor::db_object<LootGroup>
+{
+    Group m_group;
+
+    unsigned int index() const { return m_group.m_loots_map.GetMaxKey() != 0 ? m_group.m_index : 0; }
+
+    void IsValueAllowed(LootInfo current) const;
```
lafreak, line 64: Signarue suggests the return value should be true/false, but is void

apps/gameserver/include/Loot.h
```diff
@@ -0,0 +1,83 @@
...
+    void AssignGroup(std::vector<uint32_t> values_from_bracket);
+
+    virtual void set(bango::processor::lisp::var param) override;
+};
+
+
+struct LootItemGroup : public bango::processor::db_object<LootItemGroup>
```
lafreak, line 71: Should the structures in this file be structs? Do we want to expose all fields and methods?

core/utils/test/main.cpp
```diff
@@ -33,6 +34,43 @@ TEST(ClockTest, DISABLED_Difference)
...
+    interval_map<std::uint32_t> im;
+
+    //(group (index 120) (item (100 48 0) (200 1318 0) (350 48 0) (700 1318 0)(1000 1345 0)))
+    im.assign(0, 100, 48);
+    im.assign(100, 200, 1318);
+    im.assign(200, 350, 48);
+    im.assign(350, 700, 1318);
+    im.assign(700, 1000, 1345);
+
+    EXPECT_EQ(48, im[0]);
```
lafreak, line 48: Gtest requires expected value to be second parameter, actual value first. This is important to get correct failure log.

core/utils/test/main.cpp
```diff
@@ -33,6 +34,43 @@ TEST(ClockTest, DISABLED_Difference)
...
+TEST(IntervalMap, AssigningElements)
+{
+    interval_map<std::uint32_t> im;
+
+    //(group (index 120) (item (100 48 0) (200 1318 0) (350 48 0) (700 1318 0)(1000 1345 0)))
+    im.assign(0, 100, 48);
+    im.assign(100, 200, 1318);
+    im.assign(200, 350, 48);
+    im.assign(350, 700, 1318);
+    im.assign(700, 1000, 1345);
```
lafreak, line 46: Needs testing wrong assignments such as:
im.assign(0, 5, 99);
im.assign(3, 8, 99);

lafreak: What should we even expect in this scenario?

lafreak:
Rename PR to better represent the changes - "Loot parsing utilities"

apps/gameserver/src/Loot.cpp
```diff
@@ -0,0 +1,116 @@
...
+        }
+        case A_ITEM:
+        {
+            while (!param.null()) 
+            {
+                auto values_from_bracket = GetValuesFromBrackets(param);
+                AssignGroup(values_from_bracket);
+            }
+            break;
+        }
```
lafreak, line 27:
```suggestion
        case IC_MONEY:
        case A_ITEM:
        {
            while (!param.null()) 
            {
                auto values_from_bracket = GetValuesFromBrackets(param);
                AssignGroup(values_from_bracket);
            }
            break;
        }
```

Both cases are indentical, merge to one?

apps/gameserver/include/Loot.h
lafreak: I see, it checks whether the group is valid. How about rename to `Validate`?

apps/gameserver/src/Loot.cpp
```diff
@@ -0,0 +1,116 @@
...
+        loot_info = LootInfo(values_from_bracket.at(1), values_from_bracket.at(0), 1, values_from_bracket.at(2));
+    else
+        throw std::runtime_error("Invalid amount of values in brackets for group index: " + std::to_string(m_group.m_index));
+
+    IsValueAllowed(loot_info);
+    map.assign(max_key, values_from_bracket.at(0), loot_info);
+ }
+
+
+void LootItemGroup::IsValueAllowed(GroupInfo current) const
```
lafreak, line 101: We should catch those exceptions and report an error or handle the situation.

apps/gameserver/include/Loot.h
Cr3ativeZen: done

core/processor/include/bango/processor/db.h
```diff
@@ -76,7 +76,7 @@ namespace bango { namespace processor {
 
                     // BUG: Some configs have multiple types of rows for example InitNPC npc/gennpc. Filter?
                     if (m_db.find(temp.index()) == m_db.end()) 
-                        m_db.insert(std::make_pair(temp.index(), std::unique_ptr<T>(new T(temp) )));
+                        m_db.insert(std::make_pair(temp.index(), std::unique_ptr<T>(std::make_unique<T>(temp) )));
```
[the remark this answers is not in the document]
Cr3ativeZen: done

core/utils/test/main.cpp
Cr3ativeZen: done

core/utils/test/main.cpp
Cr3ativeZen: in this case we should get 99 for interval of [0,8).
I added a lot of tests which will cover edge cases.

apps/gameserver/src/Loot.cpp
Cr3ativeZen: true, done.

apps/gameserver/include/Loot.h
Cr3ativeZen: I think LootInfo, GroupInfo, Group, ItemGroup should stay as they are since they are just storing values and have some default constructors/ == operators.
I changed LootGroup and LootItemGroup to be classes and added getters.

apps/gameserver/include/Loot.h
```diff
@@ -0,0 +1,89 @@
...
+
+struct ItemGroup
+{
+    ItemGroup() = default;
+    std::uint32_t m_index;
+    bango::utils::interval_map<GroupInfo> m_groups_map;
+};
+
+
+class LootGroup : public bango::processor::db_object<LootGroup>
```
lafreak, line 58: What's the point of changing from struct to class if we expose the group field anyway? I'd rather have all config structures declared the same way instead of mixing class/struct

apps/gameserver/src/Loot.cpp
```diff
@@ -0,0 +1,122 @@
...
+            while (!param.null()) 
+            {
+                auto values_from_bracket = GetValuesFromBrackets(param);
+                try
+                {
+                    AssignGroup(values_from_bracket);
+                }
+                catch(const std::exception& e)
+                {
+                    std::cerr << e.what() << '\n';
```
lafreak, line 22: Do we want to stop reading, report an error and close the app?

lafreak: I think that is the behavior of original kal server?

core/utils/include/bango/utils/interval_map.h
```diff
@@ -0,0 +1,67 @@
...
+#include <iostream>
+
+namespace bango::utils
+{
+template<typename V>
+class interval_map
+{
+public:
+	V m_val_begin;
+	std::map<std::uint32_t, V> m_map;
```
lafreak, line 14: Do we need std::map or std::unordered_map is sufficient?

core/utils/include/bango/utils/interval_map.h
```diff
@@ -0,0 +1,67 @@
...
+    std::uint32_t max;
```
lafreak, line 15: Indent error

core/utils/include/bango/utils/interval_map.h
```diff
@@ -0,0 +1,67 @@
+#include <assert.h>
+#include <map>
+#include <limits>
+#include <cstdint>
+#include <iostream>
+
+namespace bango::utils
+{
+template<typename V>
+class interval_map
```
lafreak, line 10: Why is it a class if we make everything public anyway? Would use struct instead.

On the other hand I also think that you do expose too much internals for the user. Is it for unit testing purposes? We should only unit test the contract that this utility provides, internal stuff should be hidden.

core/utils/include/bango/utils/interval_map.h
```diff
@@ -0,0 +1,67 @@
...
+			auto [iterator, success] = m_map.insert_or_assign(keyEnd, currentKeyEndValue);
+            max = std::max(max, keyEnd);
+			endRemoveIterator = iterator;
+		}
+
+		if (beginRemoveIterator != m_map.end() && (endRemoveIterator == m_map.end() || beginRemoveIterator->first < endRemoveIterator->first))
+			m_map.erase(beginRemoveIterator, endRemoveIterator);
+	}
+
+    std::uint32_t GetMaxKey() const
```
lafreak, line 51: indent error

core/utils/test/main.cpp
```diff
@@ -33,6 +34,354 @@ TEST(ClockTest, DISABLED_Difference)
...
+}
+
+TEST(interval_map, EmptyMapTest)
+{
+	interval_map<std::uint32_t> m;
+
+	EXPECT_EQ(m[10], 0);
+
+	EXPECT_EQ(m.m_map.size(), 0);
+	IsMapCanonic(m);
```
lafreak, line 61: I don't think we need to check IsMapCanonic in all the tests, can't we make one test that checks the cannonicness?

core/utils/test/main.cpp
```diff
@@ -33,6 +34,354 @@ TEST(ClockTest, DISABLED_Difference)
...
+}
+
+TEST(interval_map, TryToAddTheSameFirstMapElementAs_m_valBegin)
+{
+	interval_map<std::uint32_t> m;
+	m.assign(5, 10, 0);
+
+	EXPECT_EQ(m.m_map.size(), 0);
```
lafreak, line 69: Instead of exposing `m_map`, you can create getter: `m.get_size()`;

lafreak: I'm also not sure what size means in this context. Is it number of ranges? If so, why is it 0 if we clearly have one range [5,10]?

core/utils/include/bango/utils/interval_map.h
Cr3ativeZen: interval_map requires std::map because upper_bound are only available for std::map.

core/utils/include/bango/utils/interval_map.h
Cr3ativeZen: In my visual studio code it's correct. Seems like bug again

core/utils/test/main.cpp
Cr3ativeZen: as stated in testcase name: TryToAddTheSameFirstMapElementAs_m_valBegin

We are trying to add an element of range [5,10) which is the same value as valbegin.
When the interval_map is empty no matter what you trying to access by [] operator will return valbegin.
When you try to add any range with value == valbegin it's pointless when map is empty.

core/utils/include/bango/utils/interval_map.h
Cr3ativeZen: same as above, buggy github

core/utils/test/main.cpp
Cr3ativeZen: We need to check if map is Canonic in every test case to make sure that assigns were done correctly.

core/utils/include/bango/utils/interval_map.h
Cr3ativeZen: Everything is hidden under private now. I added a const getter for map access only.

lafreak (approved):

lafreak (approved):

PR #32, opened 2023-02-28
Roll for drops, random::between changed, additional checks for config…

Cr3ativeZen:
… loads


apps/gameserver/include/Loot.h
```diff
@@ -63,6 +63,7 @@ struct LootGroup : public bango::processor::db_object<LootGroup>
 
     void ValidateLootInfo(LootInfo current) const;
     void AssignGroup(std::vector<uint32_t> values_from_bracket);
+    LootInfo RollLoot() const;
```
lafreak, line 66: Line 62 - makes no sense to check for 0

apps/gameserver/src/Loot.cpp
```diff
@@ -60,6 +62,9 @@ void LootItemGroup::AssignItemGroup(std::vector<uint32_t> values_from_bracket)
     if(values_from_bracket.size() != 2)
         throw std::runtime_error("Invalid amount of values in brackets for itemgroup index: " + std::to_string(m_itemgroup.m_index));
 
+    if(!LootGroup::Find(values_from_bracket.at(1)))
```
lafreak, line 65: Make the variables more descriptive. It's hard to deduce what values from bracket really mean. Assigning values_from_bracket.at(1) to temp variable with proper name would suggest more for the reader

lafreak: Probably
```
auto group_index = values_from_bracket.at(1);
```
?

apps/gameserver/src/Loot.cpp
```diff
@@ -106,3 +111,14 @@ void LootGroup::ValidateLootInfo(LootInfo current) const
         throw std::runtime_error("Chance value is lower than previous chance value for loot index: " + std::to_string(current.m_index) + " in group:" + std::to_string(m_group.m_index));
 }
 
+LootInfo LootGroup::RollLoot() const
+{
+    return m_group.m_loots_map[bango::utils::random::between(0, 1000)];
```
lafreak, line 116: Is between inclusive from both ends? If so, you've got 1001 tickets here, instead of 1000.

Is it:
[0, 1000] ? => 1001
(0, 1000] ? => 1000
[0, 1000) ? => 1000
(0, 1000) ? => 999

By looking at inix code I see that it should be 1-1000 ((0,1000])

apps/gameserver/src/Monster.cpp
```diff
@@ -65,7 +64,12 @@ void InitMonster::set(lisp::var param)
                             break;
         case A_ITEMGROUP:
         {
-            m_itemgroups.push_back(MonsterItemGroup(param.pop(), param.pop()));
+            std::uint32_t itemgroup_index = param.pop();
+            std::uint32_t rolls = param.pop();
```
lafreak, line 68: how about changing variable name to something more descriptive, not only here but in each place to:
`number_of_rolls`

rolls name feels like it actually represents already rolled results

apps/gameserver/src/Monster.cpp
```diff
@@ -175,11 +179,55 @@ void Monster::DistributeExp()
 
     }
     // TODO: Distribute party container EXP.
+
+    // TODO: Decide which player(party) will recieve loot.
+    RollLoot();
```
lafreak, line 184: I don't think this is good place to use RollLoot. Not all monsters will distribute exp probably? Why not to add similar method to `DistributeExp`? `DistributedLoot` would be good fit. It could be called in Monster::Die

apps/gameserver/src/Monster.cpp
```diff
@@ -175,11 +179,55 @@ void Monster::DistributeExp()
...
+{
+    std::vector<LootInfo> loot_vec;
+
+    for(const auto& itemgroup : m_init->m_itemgroups)
+    {
+        auto* loot_itemgroup = LootItemGroup::Find(itemgroup.m_index);
+
+        if (!loot_itemgroup)
+        {
+            spdlog::error("ItemGroup of index {} not found.", itemgroup.m_index);
```
lafreak, line 205: Can it happen? Arent we validating such configuration at server loading time?

core/utils/src/random.cpp
```diff
@@ -1,16 +1,13 @@
...
-void random::init()
-{
-    srand(time(0));
-}
-
 unsigned int random::between(unsigned int min, unsigned int max)
 {
-    return (rand() % (max+1 - min)) + min;
+    static std::random_device rd;
+    static std::mt19937 gen(rd());
```
lafreak, line 10: Please ensure that this seeds the number generator only once. Otherwise we could have repeated drops.

apps/gameserver/include/Loot.h
Cr3ativeZen: It does actually, unfortunatlly parser does scan both itemgroups and groups, which has indexes as a common part. For example when parser parse itemgroups it might parse a group aswell. when GetMaxKey() == 0 means that theres an additional group existing in this container but it has an empty interval_map.

Cr3ativeZen: To fix it, changes in load method are required which is well... tough

apps/gameserver/src/Loot.cpp
Cr3ativeZen: done

apps/gameserver/src/Loot.cpp
Cr3ativeZen: its inclusive on both ends. i changed it to 1-1000

apps/gameserver/src/Monster.cpp
Cr3ativeZen: done

apps/gameserver/src/Monster.cpp
Cr3ativeZen: true, i forget to delete this if

core/utils/src/random.cpp
Cr3ativeZen: static keyword are taking care of it.

apps/gameserver/include/Loot.h
```diff
@@ -72,10 +73,11 @@ struct LootItemGroup : public bango::processor::db_object<LootItemGroup>
 {
     ItemGroup m_itemgroup;
```
lafreak, line 74: LootItemGroup struct contains ItemGroup. Why? Wouldnt it be better to merge both classes into one?

```
struct ItemGroup : public bango::processor::db_object<LootItemGroup> {
  std::uint32_t m_index;
  bango::utils::interval_map<GroupInfo> m_groups_map;

  unsigned int index() const { return m_itemgroup.m_index; }

    void ValidateGroupInfo(GroupInfo current) const;
    void AssignItemGroup(std::vector<uint32_t> values_from_bracket);
    const LootGroup* RollGroup() const;

    virtual void set(bango::processor::lisp::var param) override;
}
```

It is very straightforward to call ItemGroup::Find(...) to find itemgroup instead of calling LootItemGroup::Find(...).
Lets name things as they appear in configuration files. We already to it for other dbs, such as: InitItem::Find(...).

lafreak: The same comment applies to LootGroup => Group

apps/gameserver/src/Loot.cpp
```diff
@@ -106,3 +117,14 @@ void LootGroup::ValidateLootInfo(LootInfo current) const
...
+LootInfo LootGroup::RollLoot() const
+{
+    return m_group.m_loots_map[bango::utils::random::between(1, 1000)];
+}
+
+const LootGroup* LootItemGroup::RollGroup() const
+{
+    auto group_index = m_itemgroup.m_groups_map[bango::utils::random::between(1, 1000)].m_index;
+    auto* group = LootGroup::Find(group_index);
+    return group != nullptr ? group : nullptr;
```
lafreak, line 129: this line is useless, just return the group - effect will be the same

apps/gameserver/src/Monster.cpp
```diff
@@ -175,11 +179,53 @@ void Monster::DistributeExp()
...
+    for(const auto& itemgroup : m_init->m_itemgroups)
+    {
+        auto* loot_itemgroup = LootItemGroup::Find(itemgroup.m_index);
+
+        for(int i = 0; i < itemgroup.m_number_of_rolls; i++)
+        {
+            auto* rolled_group = loot_itemgroup->RollGroup();
+
+            // If roll was not successful it returns index of 0.
+            if(rolled_group->index() == 0)
```
lafreak, line 212: what does "not successful" mean?
isnt better to return nullptr when it is not successful?

apps/gameserver/src/Monster.cpp
```diff
@@ -175,11 +179,53 @@ void Monster::DistributeExp()
...
+    DistributeLoot();
+}
+
+std::vector<LootInfo> Monster::RollLoot()
+{
+    std::vector<LootInfo> loot_vec;
+
+    for(const auto& itemgroup : m_init->m_itemgroups)
+    {
+        auto* loot_itemgroup = LootItemGroup::Find(itemgroup.m_index);
```
lafreak, line 205: check for possible nullptr

lafreak: The same you do in line 217

apps/gameserver/src/main.cpp
```diff
@@ -53,16 +52,14 @@ int main(int argc, char** argv)
...
-
     try
     {
         InitItem        ::Load("Config/InitItem.txt");
         InitNPC         ::Load("Config/InitNPC.txt");
-        InitMonster     ::Load("Config/InitMonster.txt");
         GenMonster      ::Load("Config/GenMonster.txt");
-        LootGroup       ::Load("Config/ItemGroup.txt");
-        LootItemGroup   ::Load("Config/ItemGroup.txt");
+        LootGroup       ::Load("Config/ItemGroup.txt", "(group");
```
lafreak, line 60: It's more userfriendly to pass "group" instead of "(group".
By user I mean us, using the API in future PRs.

To achieve it you would need to strip the name in ::Load from `(` and `)`

lafreak: Please add unit tests for that in core/processor/db

core/processor/include/bango/processor/db.h
```diff
@@ -6,14 +6,16 @@
...
+#include <cstring>
 
 namespace bango { namespace processor {
 
     template<typename T>
     class db_object
     {
     public:
-        static bool Load(const char* path) { return container::instance().load(path); }
+        static bool Load(const char* path, std::optional<const char*> name_filter = std::nullopt) { return container::instance().load(path, name_filter); }
```
lafreak, line 18: You asked for "new fashion" way of passing optional string.
But you mix new cpp std::optional with old C style `char*`. Why so?

1. It would be better to use std::optional<std::string> instead.
2. It would be even better to just use std::string. And have "" as defaut. Just check if the string is empty in Load implementation.

core/processor/include/bango/processor/db.h
```diff
@@ -69,6 +71,9 @@ namespace bango { namespace processor {
 
                     auto name = (const char*) param.pop();
 
+                    if(name_filter.has_value() && std::strcmp(name, name_filter.value()) == 0)
```
lafreak, line 74: If we switch to modern std::string we can use .equals instead of C strcmp

lafreak (approved):

PR #33, opened 2023-02-28
Enable warnings

Cr3ativeZen:


apps/gameserver/src/Spawn.cpp
```diff
@@ -40,7 +40,7 @@ void GenMonster::set(bango::processor::lisp::var param)
 
 void Spawn::Tick()
 {
-    if((time::now() - m_next_spawn_cycle).count() < GetSpawnCycle())
+    if((time::now() - m_next_spawn_cycle).count() < static_cast<std::uint64_t>(GetSpawnCycle()))
```
lafreak, line 43: What if GetSpawnCycle returns negative value?

core/network/include/bango/network/packet.h
```diff
@@ -160,7 +160,7 @@ namespace bango { namespace network {
     template<typename T>
     inline T packet::pop()
     {
-        if (m_end-m_begin < sizeof(T))
+        if (static_cast<std::uint64_t>(m_end-m_begin) < sizeof(T))
```
lafreak, line 163: Why do we need to cast? Aren't we sure that end is larger than begin?

core/persistence/src/query.cpp
```diff
@@ -4,7 +4,7 @@ namespace bango { namespace persistence {
 
     query::query(Connection_T conn, const char* q)
     {
-        m_pstatement = Connection_prepareStatement(conn, q);
+        m_pstatement = Connection_prepareStatement(conn, "%s", q);
```
lafreak, line 7: Wasn't that a bug? What is the changed behavior now?

core/space/include/bango/space/quadtree.h
```diff
@@ -37,13 +37,13 @@ namespace bango { namespace space {
         int m_x, m_y;
         unsigned char m_type;
         //! \return Distance between self and 2D point given as parameter.
-        int distance(point p) const { return distance(p.x, p.y); }
+        std::uint32_t distance(point p) const { return distance(p.x, p.y); }
```
lafreak, line 40: Entire utility bases on ints, as you can see in lines 27-30, 22, 37. What's the point of changing `distance` to unsigned?

lafreak: https://stackoverflow.com/questions/18795453/why-prefer-signed-over-unsigned-in-c

apps/gameserver/include/World.h
```diff
@@ -41,7 +41,7 @@ class WorldMap
     };
 
 private:
-    const int m_sight;
+    const std::uint32_t m_sight;
```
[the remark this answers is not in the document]
Cr3ativeZen: nvm, std::cmp_equal
std::cmp_not_equal
std::cmp_less
std::cmp_greater
std::cmp_less_equal
std::cmp_greater_equal

are the way to go

lafreak (approved):

PR #36, opened 2023-03-14
Loot part3 - Monster drops loot on the ground

Cr3ativeZen:


apps/gameserver/include/Loot.h
```diff
@@ -77,3 +78,29 @@ struct ItemGroup : public bango::processor::db_object<ItemGroup>
 };
 
 std::vector<std::uint32_t> GetValuesFromBrackets(bango::processor::lisp::var& param);
+
+
+class Loot : public Character
+{
+    ITEMINFO m_item_info;
```
lafreak, line 85: ITEMINFO might not have default constructor which initializes the data. This means we have garbage upon Loot creation. 

In Loot constructor we only fill those:
```
    m_item_info.Index = loot_info.m_index;
    m_item_info.Num = loot_info.m_amount;
    m_item_info.Prefix = loot_info.m_prefix;
```

My suggestion is to memset all zeros m_item_info first in constructor. Please look it up and decide what's most clean option.

apps/gameserver/src/Loot.cpp
```diff
@@ -136,4 +139,42 @@ const Group* ItemGroup::RollGroup() const
     }
 
     return group;
+}
+
+Loot::Loot(LootInfo loot_info, int x, int y, int map)
+    : Character(Character::LOOT)
+{
+    AssignNewId();
```
lafreak, line 147: Other Character inherited class do not need to call this in constructor. Please use the same approach.

apps/gameserver/src/Loot.cpp
```diff
@@ -136,4 +139,42 @@ const Group* ItemGroup::RollGroup() const
...
+    SetAppearTime(time::now());
```
lafreak, line 148: We do not need an API to set any time. We can encapsulate the behavior to `ResetAppearTime()` and internal implementation would choose time::now()

lafreak: It's good to never expose more functionalities than you need

apps/gameserver/include/World.h
```diff
@@ -238,6 +260,10 @@ class World
             case Character::NPC:
                 Get().m_npcs.insert(std::make_pair(entity->GetID(), std::dynamic_pointer_cast<NPC>(entity)));
                 break;
+            case Character::LOOT:
+                Get().m_loots.insert(std::make_pair(entity->GetID(), std::dynamic_pointer_cast<Loot>(entity)));
+                auto loots = Get().m_loots;
```
lafreak, line 265: ?

apps/gameserver/src/Loot.cpp
```diff
@@ -1,8 +1,11 @@
 #include "Loot.h"
 
 #include <bango/utils/random.h>
+#include "World.h"
```
lafreak, line 4: Why is World needed?

apps/gameserver/src/World.cpp
```diff
@@ -99,6 +99,14 @@ void WorldMap::Add(Character* entity)
             if (monster->distance(center) <= m_sight)
                 m_on_appear(*(Player*) entity, *monster, false);
         }
+
+        for (auto& p : container->loots())
+        {
+            auto loot = p.second;
+
+            if (loot->distance(center) <= m_sight)
+                m_on_appear(*(Player*) entity, *loot, false);
```
lafreak, line 108: Missing `m_on_disappear` call for loot
World.cpp L187-L200

apps/gameserver/src/Loot.cpp
Cr3ativeZen: ups, done

apps/gameserver/include/World.h
Cr3ativeZen: ups, done

apps/gameserver/src/Loot.cpp
Cr3ativeZen: forget to remove, done

apps/gameserver/src/World.cpp
Cr3ativeZen: done

apps/gameserver/src/Loot.cpp
Cr3ativeZen: done

apps/gameserver/include/Loot.h
Cr3ativeZen: memset added.

lafreak (approved):

PR #39, opened 2023-03-18
Loot is pickable now- loot saves in DB, few minor fixes

Cr3ativeZen:


apps/gameserver/include/Loot.h
```diff
@@ -87,12 +87,15 @@ class Loot : public Character
 public:
     static constexpr std::uint32_t DISAPPEAR_TIME = 180000;
     static constexpr std::uint32_t PRIORITY_TIME = 120000;
+    static constexpr std::int32_t MAX_DISTANCE_FROM_TARGET = 20;
+    static constexpr std::int32_t MIN_DISTANCE_FROM_TARGET = -20;
```
lafreak, line 91: The naming is wrong, distance cannot be below 0, how can it be?
I think you wanted to create 1 constant for `MAX_DISTANCE_FROM_TARGET = 20;` and then use it in randomization:
```
random(x-MAX_DISTANCE_FROM_TARGET , y+MAX_DISTANCE_FROM_TARGET )
```

lafreak: This would not be distance as well, distance would be the circle with `MAX_DISTANCE_FROM_TARGET ` radius, but it does not really matter

apps/gameserver/src/Loot.cpp
```diff
@@ -144,8 +144,8 @@ Loot::Loot(LootInfo loot_info, int x, int y, int map)
     : Character(Character::LOOT)
 {
     ResetAppearTime();
-    m_x = x;
-    m_y = y;
+    m_x = x + random::between_signed(MIN_DISTANCE_FROM_TARGET, MAX_DISTANCE_FROM_TARGET);
```
lafreak, line 147: Loot constructor gets x and y information. We should assume that this is expected loot's x and y. But it looks like its dropper x/y which is misleading.

I think it would be better to take out dropper X/Y info from Loot constructor and let it be drop x&y like in all other Characters.

To achieve it you could move x/y randomization one call above, before initializing new Loot.

core/utils/src/random.cpp
```diff
@@ -10,4 +10,12 @@ unsigned int random::between(unsigned int min, unsigned int max)
     static std::mt19937 gen(rd());
     std::uniform_int_distribution<unsigned int> distr(min, max);
     return distr(gen);
+}
+
+int random::between_signed(int min, int max)
+{
+    static std::random_device rd;
+    static std::mt19937 gen(rd());
+    std::uniform_int_distribution<int> distr(min, max);
```
lafreak, line 19: This call can be templatized:

```
template <typename T>
T random::between(T min, T max)
{
    static std::random_device rd;
    static std::mt19937 gen(rd());
    std::uniform_int_distribution<T> distr(min, max);
    return distr(gen);
}
```

apps/gameserver/include/Loot.h
Cr3ativeZen: good point, done

apps/gameserver/src/Loot.cpp
Cr3ativeZen: done

core/utils/src/random.cpp
Cr3ativeZen: there were some issues with templated std::uniform_int_distribution. I left the between method as it was before.

apps/gameserver/include/Loot.h
```diff
@@ -87,12 +87,14 @@ class Loot : public Character
 public:
     static constexpr std::uint32_t DISAPPEAR_TIME = 180000;
     static constexpr std::uint32_t PRIORITY_TIME = 120000;
+    static constexpr std::int32_t MAX_DISTANCE_FROM_TARGET = 20;
```
lafreak, line 90:
```suggestion
    static constexpr std::int32_t MAX_DISTANCE_FROM_THROWER = 20;
```
This is no to mistake from appear sight distance

lafreak (approved):

PR #49, opened 2023-03-25
[Lvl 101 Bug Fix] Checks to Player::OnAttack added to prevent Dying multiple times

Cr3ativeZen:


apps/gameserver/src/Player.cpp
```diff
@@ -692,7 +692,7 @@ void Player::OnAttack(packet& p)
 
         character.ReceiveDamage(GetID(), damage);
 
-        if (character.GetCurHP() <= 0)
+        if (character.GetCurHP() <= 0 && !character.IsGState(CGS_KNEE) && !character.IsGState(CGS_KO))
```
lafreak, line 695: We should check for monster HP or KO/KNEE state right after we receive lock. This way we are sure that nobody will modify monster together with us, so it is thread safe. We can then decide to abort the attack if condition is met.

lafreak:
Outdated

PR #51, opened 2023-03-25
Update Player.cpp

ArslaAnN:


apps/gameserver/src/Player.cpp
```diff
@@ -925,3 +925,8 @@ void Player::LevelUp()
     ApplyVisualEffect(E_LEVELUP);
     //m_data.Exp = 0;
 }
+
+void Player::PrivateNotice (std::string message)
```
lafreak, line 929:
```suggestion
void Player::PrivateNotice (const std::string& message)
```
In performance critical paths its better not to copy string message each time function is called. To pass string by reference, without making copy, use `&`.

Additionally, use `const` to enforce the function not to modify original object.

Read further: https://faculty.cs.niu.edu/~mcmahon/CS241/Notes/passing_by_reference.html

ArslaAnN: alright, thank you <3

PR #54, opened 2023-03-31
Skills learning and saving them to database. Loading skills OnLogin

Cr3ativeZen:


apps/shared/inix/protocol.h
```diff
@@ -246,6 +246,8 @@ enum S2D_PROTOCOL
 	//S2D_PUTOFFITEM,
 	S2D_UPDATEITEMINFO,
 	S2D_SHORTCUT,
+	S2D_SKILLUP,
+	S2D_INSERTSKILL
```
lafreak, line 250: Minor:
InsertNewSkil / InsertSkill / INSERTSKILL can be replaced with LearnSkill / LEARNSKILL to match inix packet/method names

apps/dbserver/src/DatabaseManager.cpp
```diff
@@ -681,4 +715,43 @@ void DatabaseManager::SaveAllProperty(const std::shared_ptr<GameServer>& s, pack
...
+    auto query2 = conn.create_query("UPDATE player SET supoint=? WHERE idplayer=?");
+
+    query2 << skill_points << id_player;
+    query2.execute();
+}
+
+
+void DatabaseManager::UpgradeSkill(const std::shared_ptr<GameServer>& s, packet& p)
+{
+    auto id_player = p.pop<id_t>();
```
lafreak, line 742: Do not use `id_t` for idplayer, it can be different than PID in future. `id_t` is reserved for on map entities with self-assigned temporary id - identifier for map.
Use `unsigned int` like in LoadPlayers

apps/gameserver/include/Skill.h
```diff
@@ -1,16 +1,17 @@
...
 #include <inix.h>
+#include <bango/network/packet.h>
 
 class InitSkill : public bango::processor::db_object<InitSkill>
 {
+    static constexpr std::uint8_t MAX_SKILL_INDEX = 84;
     enum Kind : int64_t { ANIMAL=1, MONSTER=2 };
 
 public:
-    unsigned int index() const;
```
lafreak, line 11: Why moving it below? It nicely separates overrides (which are not part of interface for InitSkill class users) from utility functions (API) - Find...Skill

apps/gameserver/include/Skill.h
```diff
@@ -31,4 +32,22 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+};
+
+
+class Skill
+{
+    const InitSkill* m_init;
+    std::uint32_t m_last_use;
+    std::uint8_t m_level;
+
+    public:
```
lafreak, line 47: indent

apps/gameserver/include/Skill.h
```diff
@@ -31,4 +32,22 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+    Skill(const InitSkill* init,const std::uint8_t skill_level);
+
+    const std::uint8_t GetLevel() const { return m_level; }
+    const std::uint8_t GetIndex() const { return m_init->Index; }
+    void LevelUp() { ++m_level; }
```
lafreak, line 52: This function does not give control to the caller what is the grade he wants the skill to be. Since it needs to match skill manager level, we should call explicitly what level we want it to be. Therefore I assume we need different method - `SetLevel`

apps/gameserver/include/Skill.h
```diff
@@ -31,4 +32,22 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
     static InitSkill* FindMonsterSkill(std::uint8_t skill_index);
+    unsigned int index() const;
+    virtual void set(bango::processor::lisp::var param) override;
+
+};
+
+
+class Skill
+{
+    const InitSkill* m_init;
```
lafreak, line 43: We had convention of keeping reference to unique ptr config object. Check other places. Why do we want to change now? If so, why only here?

If we want to go with this approach, you made the `m_init` variable const, not the data underneath. Whats the difference?

- you cannot change m_init ptr to something else (`m_init = other_ptr` will not work) and that is right
- you CAN change data underneath, for example `m_init->MaxLevel=5;` will compile which is not desired

Another question to this approach, why do we need pointer here? If we are sure the initskill is never unloaded and always exists, we can use references

apps/gameserver/include/SkillManager.h
```diff
@@ -0,0 +1,19 @@
...
+#include <unordered_map>
+#include <memory>
+
+#include "Skill.h"
+
+#include <bango/network/packet.h>
+
+class SkillManager
+{
+    public:
```
lafreak, line 12: indent, please check all other places as well

apps/gameserver/include/SkillManager.h
```diff
@@ -0,0 +1,19 @@
...
+    std::unordered_map<std::uint8_t, std::unique_ptr<Skill>> m_skills;
+
+    void UpgradeSkill(const std::uint8_t skill_id);
```
lafreak, line 15: Should upgrade skill method return boolean with information if upgrade succeeded?

apps/gameserver/src/DBListener.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "Socket.h"
 #include "Player.h"
+#include <iostream>
```
lafreak, line 7: not required

apps/gameserver/src/Player.cpp
```diff
@@ -18,6 +18,7 @@ using namespace bango::utils;
 
 Player::Player(const bango::network::taco_client_t& client) : User(client), Character(Character::PLAYER)
 {
+    //m_skills(AssignSkillsToPlayer)
```
lafreak, line 21: ?

apps/gameserver/src/Player.cpp
```diff
@@ -112,6 +109,34 @@ void Player::OnLoadItems(packet& p)
...
+        if(!skill_init)
+        {
+            spdlog::warn("Player {} has an invalid skill id {} in database", GetName(), skill_id);
+            continue;
+        }
+
+        spdlog::info("Player {} loaded skill {} at grade {}", GetName(), skill_id, skill_level);
+        InsertSkill(skill_init, skill_id, skill_level);
+
+        std::uint32_t cooldown_remaining = 0;
```
lafreak, line 132: Add todo

apps/gameserver/src/Player.cpp
```diff
@@ -811,6 +836,103 @@ void Player::OnItemPick(packet& p)
...
+        spdlog::warn("Player {} tried to learn an invalid skill id {}", GetName(), skill_id);
+        return false;
+    }
+
+    if(skill_init->LevelLimit > GetLevel())
+    {
+        spdlog::warn("Player {} tried to learn a skill id {} but is not high enough level", GetName(), skill_id);
+        return false;
+    }
+    if(skill_init->Job > GetJob())
```
lafreak, line 900: Job is a mask, not a number. For example, basic job is `1`, first job is `2`, second job vaga is `4` and comma is `8`.
This is why vaga is 7 - `1+2+4` and comma 11 - `1+2+8` and jobhack is 15: `1+2+4+8.

This check should not check equality but existence of flag.

apps/gameserver/src/Player.cpp
```diff
@@ -811,6 +836,103 @@ void Player::OnItemPick(packet& p)
...
+        return false;
+    }
+
+    if(HasSkill(skill_id) && GetSkill(skill_id)->GetLevel() >= skill_init->MaxLevel)
+    {
+        spdlog::warn("Player {} tried to learn a skill id {} but it is already max level", GetName(), skill_id);
+        return false;
+    }
+
+    const auto required_skill_id = skill_init->RequiredSkillId;
```
lafreak, line 917: RequiredSkillIndex fits better

apps/gameserver/src/Player.cpp
```diff
@@ -880,6 +1002,14 @@ void Player::Die()
     WriteInSight(bango::network::packet(S2C_ACTION, "db", GetID(), AT_DIE));
 }
 
+void Player::ResetStates()
+{
+    Character::ResetStates();
+    m_inventory = Inventory();
+    LeaveParty();
+    m_skills.clear();
```
lafreak, line 1010: I think that skill manager should be done similarly to Inventory, instead of inheritance use member `m_skill_manager` field. This way we can hide m_skills map from the `Player` class.

```suggestion
    m_skills = SkillManager();
```

apps/gameserver/include/SkillManager.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
lafreak, line 1: To keep the current convention I think we can keep the manager in Skill header/cpp unit

lafreak: In the future we need to split all classes into different cpp units to decrease compile time

apps/gameserver/src/Skill.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include <inix.h>
 
+using namespace bango::network;
```
lafreak, line 7: Not used

apps/gameserver/src/SkillManager.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+bool SkillManager::HasSkill(const std::uint8_t skill_id) const
+{
+    return m_skills.find(skill_id) != m_skills.end();
+}
+
+
+Skill* SkillManager::GetSkill(const std::uint8_t skill_id) const
+{
+    return HasSkill(skill_id) ? m_skills.at(skill_id).get() : nullptr;
```
lafreak, line 15: If this is the only reason to have Skill manager at all, I doubt we even need the manager. Maybe expose the map to the player class and thats enough...

apps/dbserver/src/DatabaseManager.cpp
Cr3ativeZen: done

apps/gameserver/include/Skill.h
Cr3ativeZen: done

apps/gameserver/include/Skill.h
Cr3ativeZen: done

apps/gameserver/include/SkillManager.h
Cr3ativeZen: done

apps/gameserver/src/DBListener.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Skill.cpp
Cr3ativeZen: done

apps/gameserver/include/SkillManager.h
Cr3ativeZen: done

apps/gameserver/src/SkillManager.cpp
Cr3ativeZen: I think it will be good to have it once we expand InsertSkill method.

apps/shared/inix/protocol.h
Cr3ativeZen: done

apps/gameserver/include/Skill.h
Cr3ativeZen: > Another question to this approach, why do we need pointer here? If we are sure the initskill is never unloaded and always exists, we can use references

In future we will make it possible to reload all kind of Init's. This might get problematic with references then.

Cr3ativeZen: Tbh i was following the convention of the return type of Find... methods. Thats why they are raw pointers.

apps/gameserver/include/Skill.h
```diff
@@ -1,10 +1,17 @@
 #pragma once
 
+#include <unordered_map>
+#include <memory>
+
+#include "Character.h"
+
 #include <bango/processor/db.h>
 #include <inix.h>
+#include <bango/network/packet.h>
```
lafreak, line 10: not needed

build.sql
```diff
@@ -55,6 +55,17 @@ CREATE TABLE shortcut (
...
+
+CREATE table skill (
+	idplayer INT NOT NULL,
+	idskill SMALLINT NOT NULL,
+	level TINYINT NOT NULL,
+
+	FOREIGN KEY (idplayer)
+		REFERENCES player (idplayer)
+);
+
```
lafreak, line 67: Default skills upon player insertion are missing (behead, run, and others) (0, 1, 11 if i remember correctly)

apps/gameserver/src/main.cpp
```diff
@@ -123,6 +125,7 @@ int main(int argc, char** argv)
     Socket::DBClient().when(D2S_ANS_NEWPLAYER,      std::bind(&DBListener::OnNewPlayerAnswer,   _1));
     Socket::DBClient().when(D2S_LOADPLAYER,         std::bind(&DBListener::OnLoadPlayer,        _1));
     Socket::DBClient().when(D2S_LOADITEMS,          std::bind(&DBListener::OnLoadItems,         _1));
+    Socket::DBClient().when(D2S_SKILLINFO,          std::bind(&DBListener::OnLoadSkills,         _1));
```
lafreak, line 128: indent

apps/dbserver/src/DatabaseManager.cpp
```diff
@@ -490,6 +498,7 @@ void DatabaseManager::LoadPlayer(const std::shared_ptr<GameServer>& s, packet& p
     s->write(out);
 
     LoadItems(s, id, idplayer);
+    LoadSkills(s, id, idplayer);
```
lafreak, line 501: Check Player::OnLoadItems. It calls OnLoadFinish which ends loading phase for player. By adding LoadSkills here you make the player loading screen start too soon. Detected by bots

build.sql
Cr3ativeZen: done

apps/gameserver/src/main.cpp
Cr3ativeZen: done

apps/dbserver/src/DatabaseManager.cpp
Cr3ativeZen: done

apps/gameserver/include/Skill.h
```diff
@@ -31,4 +37,29 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+};
+
+class SkillManager
+{
+public:
+    std::unordered_map<std::uint8_t, std::unique_ptr<Skill>> m_skills;
+
+    bool Upgrade(const std::uint8_t index, const std::uint8_t level);
+    bool Learn(const InitSkill* init,const std::uint8_t index, const std::uint8_t level = 1);
+    bool Exists(const std::uint8_t index) const;
```
lafreak, line 63: Why is it const? Its copy anyway. All these variables not need to be const (besided InitSkill)

apps/gameserver/include/Skill.h
```diff
@@ -31,4 +37,29 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+class Skill
+{
+    const InitSkill* m_init;
+    std::uint32_t m_last_use;
+    std::uint8_t m_level;
+
+public:
+    Skill(const InitSkill* init,const std::uint8_t level);
+
+    const std::uint8_t GetLevel() const { return m_level; }
```
lafreak, line 51: Return value not need to be const, the same for GetIndex

lafreak: And constructor, and other places. Please use const in places where you want to ensure calleer does not make stupid things

apps/gameserver/src/Player.cpp
```diff
@@ -811,6 +833,115 @@ void Player::OnItemPick(packet& p)
...
+void Player::LearnOrUpgradeSkill(const std::uint8_t skill_index)
+{
+    if(!CanLearnSkill(skill_index))
+        return;
+
+    const auto* init = InitSkill::FindPlayerSkill((PLAYER_CLASS)GetClass(), skill_index);
+
+
+    auto* skill = m_skills.GetByIndex(skill_index);
+    const std::uint8_t level = skill ? skill->GetLevel() + 1 : 1;
```
lafreak, line 863: Why is it const?

apps/gameserver/src/Player.cpp
```diff
@@ -811,6 +833,115 @@ void Player::OnItemPick(packet& p)
...
+    SendProperty(P_SUPOINT);
+}
+
+bool Player::CanLearnSkill(const std::uint8_t index) const
+{
+    const auto* init = InitSkill::FindPlayerSkill((PLAYER_CLASS)GetClass(), index);
+
+    if(!init)
+    {
+        spdlog::warn("Player {} tried to learn an invalid skill id {}", GetName(), index);
```
lafreak, line 900:
```suggestion
        spdlog::warn("Player {} tried to learn an invalid skill index: {}", GetName(), index);
```

apps/gameserver/src/Player.cpp
```diff
@@ -811,6 +833,115 @@ void Player::OnItemPick(packet& p)
...
+        return false;
+    }
+
+    if(init->LevelLimit > GetLevel())
+    {
+        spdlog::warn("Player {} tried to learn a skill id {} but is not high enough level", GetName(), index);
```
lafreak, line 906:
```suggestion
        spdlog::warn("Player {} tried to learn a skill index {} but is not high enough level", GetName(), index);
```
Please re-check manually all variables and logs for skill_index skill_level

apps/gameserver/src/Player.cpp
```diff
@@ -811,6 +833,115 @@ void Player::OnItemPick(packet& p)
...
+        return false;
+    }
+
+    if(m_skills.Exists(index) && m_skills.GetByIndex(index)->GetLevel() >= init->MaxLevel)
+    {
+        spdlog::warn("Player {} tried to learn a skill id {} but it is already max level", GetName(), index);
+        return false;
+    }
+
+    const auto required_skill_index = init->RequiredSkillIndex;
```
lafreak, line 926: Why do we need the variable? Just use `init->RequiredSkillIndex`

apps/gameserver/include/Skill.h
Cr3ativeZen: done

apps/gameserver/include/Skill.h
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
```diff
@@ -811,6 +833,115 @@ void Player::OnItemPick(packet& p)
...
+    if(!m_skills.Exists(index))
+    {
+        spdlog::warn("Player {} tried to upgrade skill {} without learning it first", GetName(), index);
+        return;
+    }
+    LearnOrUpgradeSkill(index);
+}
+void Player::OnSkillLearn(bango::network::packet& p)
+{
+    const auto index = p.pop<std::uint8_t>();
```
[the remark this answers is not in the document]
Cr3ativeZen: done

apps/gameserver/src/Player.cpp
```diff
@@ -811,6 +833,115 @@ void Player::OnItemPick(packet& p)
...
+        spdlog::warn("Player {} tried to learn an invalid skill id {}", GetName(), index);
+        return false;
+    }
+
+    if(init->LevelLimit > GetLevel())
+    {
+        spdlog::warn("Player {} tried to learn a skill id {} but is not high enough level", GetName(), index);
+        return false;
+    }
+    if(init->Job & GetJob())
```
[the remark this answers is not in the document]
Cr3ativeZen: done

lafreak (approved):

PR #58, opened 2023-04-01
Fix issue with receiving chunked packets

lafreak:
The server handled only packets which content is entirely contained inside one recv buffer. Now it keeps track of remaining unhandled bytes and waits for next chunk to concatenate and handle.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -39,7 +39,7 @@
                 "--db_port", "2999",
                 "--game_address", "0.0.0.0",
                 "--game_port", "3000",
-                "--processing_threads", "1",
+                "--processing_threads", "12",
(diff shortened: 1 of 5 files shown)


.vscode/launch.json
```diff
@@ -39,7 +39,7 @@
                 "--db_port", "2999",
                 "--game_address", "0.0.0.0",
                 "--game_port", "3000",
-                "--processing_threads", "12",
+                "--processing_threads", "1",
```
Cr3ativeZen, line 42: is this change intended?

lafreak: fixed

Cr3ativeZen (approved):

PR #61, opened 2023-04-10
Base Skill class extended, added subclasses,and factory method

Cr3ativeZen:


apps/gameserver/include/Skill.h
```diff
@@ -40,22 +44,62 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+    RESIST_TYPE resist_type;
 
 public:
     Skill(const InitSkill* init,std::uint8_t level);
 
     std::uint8_t GetLevel() const { return m_level; }
     std::uint8_t GetIndex() const { return m_init->Index; }
     void SetLevel(std::uint8_t level) { m_level = level; }
+
+    bango::utils::time::point GetRemainingCooldown() const { return m_last_use_time + std::chrono::milliseconds(m_init->CoolDown); }
```
lafreak, line 58: I suppose that the reason we have this method is to get miliseconds left to send to the player logging-in (S2C_SKILLINFO).
If so, currently it does not return remaining miliseconds but time when it can be used in future.

apps/gameserver/include/Skill.h
```diff
@@ -40,22 +44,62 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+    void UpdateCooldownTime() { m_last_use_time = bango::utils::time::now();}
```
lafreak, line 59:
```suggestion
    void SaveSkillUseTime() { m_last_use_time = bango::utils::time::now();}
```

apps/gameserver/include/Skill.h
```diff
@@ -40,22 +44,62 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+    virtual void Execute(const Player& player, bango::network::packet& packet) {}
+    virtual bool CanExecute(const Player& player,  const Player& target) const;
```
lafreak, line 61: CanExecute target param is of type Player, what if target is monster? I think we should have base class Character here.

apps/gameserver/src/Skill.cpp
```diff
@@ -111,16 +114,117 @@ bool SkillManager::Upgrade(std::uint8_t index, std::uint8_t level)
...
+                default:
+                {
+                    return std::make_unique<PhysicalSkill>(init, level, (ATTACK_TYPE)ATT_RANGE);
+                    break;
+                }
+            }
+        }
+        default:
+        {
+            std::runtime_error("Unknown player class while creating skill!");
```
lafreak, line 173: I think we support thief/shaman already so why not adding behead allowing that?

apps/gameserver/src/Skill.cpp
```diff
@@ -111,16 +114,117 @@ bool SkillManager::Upgrade(std::uint8_t index, std::uint8_t level)
...
+            switch(index)
+            {
+                case(SKILL_BEHEAD):
+                {
+                    return std::make_unique<Behead>(init, level);
+                    break;
+                }
+                default:
+                {
+                    return std::make_unique<PhysicalSkill>(init, level, (ATTACK_TYPE)ATT_RANGE);
```
lafreak, line 166: Default skill should be empty one, possibly `Passive`, with no passive abilities. Its entirely missing in this commit

apps/gameserver/src/Skill.cpp
```diff
@@ -111,16 +114,117 @@ bool SkillManager::Upgrade(std::uint8_t index, std::uint8_t level)
...
+                    return std::make_unique<PhysicalSkill>(init, level, (ATTACK_TYPE)ATT_MEELE);
+                    break;
+                }
+            }
+        }
+        case PLAYER_CLASS::PC_MAGE:
+        {
+            switch(index)
+            {
+                case(SKILL_BEHEAD):
```
lafreak, line 143:
```suggestion
                case SKILL_BEHEAD:
```

apps/gameserver/src/Skill.cpp
```diff
@@ -111,16 +114,117 @@ bool SkillManager::Upgrade(std::uint8_t index, std::uint8_t level)
...
+                {
+                    return std::make_unique<Behead>(init, level);
+                    break;
+                }
+                default:
+                {
+                    return std::make_unique<MagicSkill>(init, level);
```
lafreak, line 150: We should also notify that undefined skill is loaded from database (error?)

apps/gameserver/include/Skill.h
```diff
@@ -40,22 +44,62 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+    std::int32_t GetAttack() const { return 0; } // TODO: Implement
+    RESIST_TYPE GetResist() const { return resist_type; }
+    virtual bango::network::packet BuildCastPacket(const Player& player, id_t target_id = 0) const;
+};
+
+class PhysicalSkill : public Skill
+{
+    ATTACK_TYPE m_attack_type;
```
lafreak, line 69: How many bytes is `ATTACK_TYPE`?

apps/gameserver/include/Skill.h
```diff
@@ -40,22 +44,62 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
+public:
+    PhysicalSkill(const InitSkill* init, std::uint8_t level, ATTACK_TYPE attack_type)
```
lafreak, line 71: Move implementations into cpp file to speed up compile time, also consider grouping all skills into subdirectory

apps/gameserver/src/Skill.cpp
```diff
@@ -111,16 +114,117 @@ bool SkillManager::Upgrade(std::uint8_t index, std::uint8_t level)
...
+
+bool Skill::CanExecute(const Player& player, const Player& target) const
+{
+    return player.IsNormal()
+        && target.IsNormal()
+        && player.GetMap() == target.GetMap()
+        && !player.IsGState(CGS_ONTRANSFORM)
+        //&& player.IsWState(WS_WEAPON) // TODO: add this check when WearState is added.
+        && player.GetCurMP() >= m_init->MP
+        && player.CanAttack(target);
```
lafreak, line 211: Where are those checks from? This will not work for Heal skill, since 1. there is no target, 2. there is no attacking involved at all

apps/gameserver/include/Skill.h
Cr3ativeZen: done

apps/gameserver/include/Skill.h
```diff
@@ -40,22 +44,62 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
 public:
     Skill(const InitSkill* init,std::uint8_t level);
 
     std::uint8_t GetLevel() const { return m_level; }
     std::uint8_t GetIndex() const { return m_init->Index; }
     void SetLevel(std::uint8_t level) { m_level = level; }
+
+    bango::utils::time::point GetRemainingCooldown() const { return m_last_use_time + std::chrono::milliseconds(m_init->CoolDown); }
+    void UpdateCooldownTime() { m_last_use_time = bango::utils::time::now();}
+    virtual void Execute(const Player& player, bango::network::packet& packet) {}
```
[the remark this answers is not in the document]
Cr3ativeZen: done

apps/gameserver/src/Skill.cpp
Cr3ativeZen: done

apps/shared/inix/common.h
```diff
@@ -585,6 +585,8 @@ static BASEPROPERTY g_baseproperty[] = {
 #define SVMAP_TILE_COUNT			256
 #define SVMAP_TILE_SIZE				32
 
+#define SKILL_BEHEAD 1
```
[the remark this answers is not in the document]
Cr3ativeZen: done

apps/gameserver/src/Skill.cpp
Cr3ativeZen: done

apps/gameserver/src/Skill.cpp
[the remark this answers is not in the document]
Cr3ativeZen: yeah, i forget to change that. Done

apps/gameserver/src/Skill.cpp
Cr3ativeZen: for now we can add an if which will check if the skill id is less than 84.
Later on we have to make some mechanism which will check if the skill exists(not all skills 0-84 exists)

apps/gameserver/src/Skill.cpp
Cr3ativeZen: We can override and not use base CanExecute at all.
But it does not solve the problem of the Target in parameter.
Maybe std::optional?

apps/gameserver/include/Skill.h
```diff
@@ -40,22 +47,69 @@ class InitSkill : public bango::processor::db_object<InitSkill>
...
-    std::uint32_t m_last_use;
+    bango::utils::time::point m_last_use_time;
     std::uint8_t m_level;
+    RESIST_TYPE resist_type;
+protected:
+    const Player& m_player;
 
 public:
-    Skill(const InitSkill* init,std::uint8_t level);
+    Skill(const InitSkill* init, const Player& player, std::uint8_t level);
```
lafreak, line 57: Skills will most likely modify player (buffs?) so it should not be const?

apps/gameserver/include/Skill.h
```diff
@@ -40,22 +47,69 @@ class InitSkill : public bango::processor::db_object<InitSkill>
 class Skill
 {
     const InitSkill* m_init;
-    std::uint32_t m_last_use;
+    bango::utils::time::point m_last_use_time;
     std::uint8_t m_level;
+    RESIST_TYPE resist_type;
+protected:
+    const Player& m_player;
```
lafreak, line 54: Are you sure it needs to be Player? What about monster skills? Animal skills?

apps/gameserver/src/Skill.cpp
```diff
@@ -111,16 +114,122 @@ bool SkillManager::Upgrade(std::uint8_t index, std::uint8_t level)
...
+            }
+        }
+        case PLAYER_CLASS::PC_ARCHER:
+        {
+            switch(index)
+            {
+                case SA_BEHEAD:
+                    return std::make_unique<Behead>(init, m_player, level);
+                default:
+                    return std::make_unique<PhysicalSkill>(init, m_player, level, (ATTACK_TYPE)ATT_RANGE);
```
lafreak, line 153: Just use native c/cpp type instead of ATTACK_TYPE to avoid casting. Just make sure you know how many bytes does the enum have. If I remember correctly you can define it yourself:


```
enum ATT : std::uint8_t {
  ...
}
```

Cr3ativeZen: done

apps/gameserver/include/Skill.h
Cr3ativeZen: done

apps/gameserver/include/Skill.h
Cr3ativeZen: done

apps/gameserver/include/Skill.h
Cr3ativeZen: well, this method was created for future cooldown checks serverside on skill use

Cr3ativeZen: But your way seems like it can handle both of the cases.

Cr3ativeZen: done

apps/gameserver/include/Skill.h
Cr3ativeZen: sure, but its a task for 2 years from now :D

apps/gameserver/src/Skill.cpp
Cr3ativeZen: x

apps/gameserver/src/Skill.cpp
Cr3ativeZen: Passive skills should be created in different commit.

apps/gameserver/include/Skill.h
lafreak: It does matter in case you put it into network packet. Use `typedef` or using to not explicitly cast to ATTACK_TYPE in each function

apps/gameserver/include/Skill.h
```diff
@@ -40,22 +47,69 @@ class InitSkill : public bango::processor::db_object<InitSkill>
 class Skill
 {
     const InitSkill* m_init;
-    std::uint32_t m_last_use;
+    bango::utils::time::point m_last_use_time;
     std::uint8_t m_level;
+    RESIST_TYPE resist_type;
+protected:
+    Character& m_player;
```
lafreak, line 54: Why variable name is player but type Character?

apps/gameserver/include/Skill.h
lafreak: Still, I would change the name, cooldown is the static time how often you can use the skill. You do not update cooldown here, but rather save skill use time.

lafreak:
Needs rebase

apps/gameserver/include/Skill.h
Cr3ativeZen: because i copy paste, why? :D

lafreak:
Closing as duplicated, another PR #62 merged to master

